PR #193, opened 2017-05-22
Q&A - Adding answers

adamkarminski:


resources/assets/js/components/global/form/Form.vue
```diff
@@ -91,15 +89,17 @@
 								this.mutation(types.FORM_RESET)
 								this.mutation(types.FORM_UPDATE_ORIGINAL_DATA)
 							}
+
+							this.$emit('submitSuccess')
```
adamkarminski: You used event name `submitSuccess` but we decided to use `submitForm`. Please correct

PR #218, opened 2017-06-02
Layout | Navbar based on breakpoints

bkoval:
## Changelog (navbar only):
* sidenav-trigger (hamburger) shows on mobile + tablet
* controls on the right (search icon etc.) show on tablet + desktop
* breacdrumbs show on desktop
* breadcrumbs are vertically-aligned with logo

@adamkarminski @bognix

### DIFF
--- a/resources/assets/js/store/modules/ui.js
+++ b/resources/assets/js/store/modules/ui.js
@@ -3,12 +3,25 @@ import { set } from 'vue'
 
 // Initial state
 const state = {
-	currentLayout: "",
+	currentLayout: ""
+}
+
+const layouts = {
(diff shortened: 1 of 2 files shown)


resources/assets/js/store/modules/ui.js
```diff
@@ -3,12 +3,25 @@ import { set } from 'vue'
 
 // Initial state
 const state = {
-	currentLayout: "",
+	currentLayout: ""
```
adamkarminski: Minor thing, but we use `'` for strings. :)

adamkarminski:
putin-thumbs-up-262x300

PR #242, opened 2017-06-04
Slide parser fixes

gitkuba:


lib/SlideParser/Parser.php
```diff
@@ -259,9 +259,11 @@ public function getBackground($html)
 
 		$url = $match[0][1];
 
-		$image = Image::make($url)
-			->resize(1920, 1080)
-			->stream('jpg', 80);
+		$canvas = Image::canvas(1920, 1080, '#fff');
+
+		$background = Image::make($url)->resize(1920, 1080);
```
bkoval: Since canvas and background use the same values, how about moving these values to consts?

PR #304, opened 2017-06-15
User Settings | Chat uses setting to open

bkoval:
@adamkarminski 
kermit


adamkarminski:
Change the toggleChat to a universal one and LGTM

beatles-thumbs-up

PR #388, opened 2017-06-23
WIP | Final invoice

gitkuba:


resources/views/mail/final-invoice-notification.blade.php
```diff
@@ -0,0 +1,7 @@
+@extends('mail.layout')
```
gitkuba: @adamkarminski, would you please use your wonderful protein API skills on this email template? No hurry :smile:

adamkarminski: <3

PR #515, opened 2017-08-08
Notifications | Open stream notifications in a new tab

adamkarminski:
@bognix - The best solution would be to mark a notification as seen/read not at click but when you visit a page with a `&notification` param in URL. But! That would require synchronizing which I decided not to implement now, as we have a lot of work with Search.

This solution is definitely suboptimal, but IMO - good enough.

Let me know what you think. :)


bognix:
makes sense. I was checking the router-link source code if there is any event emitted "on navigate" but there is no such event :( that would be perfect

PR #524, opened 2017-08-16
WIP: Questions

adamkarminski:
:boom: __Release procedure:__ 
- [x] Set up quiz questions index !
- [x] In `elasticsearch.yml` set `indices.query.bool.max_clause_count: 10000`
- [ ] `php artisan quiz:import koniec`
- [ ] upload `taxonomies.tsv` and `php artisan import:taxonomies {filename}`
- [ ] `php artisan tags:fromTaxonomies`
- [ ] upload `exams.json` and `php artisan quiz:tagsFromMap {filename}`
- [ ] `php artisan scout:import 'App\Models\QuizQuestion'`
- [ ] `alter table user_quiz_results add key (quiz_question_id, quiz_answer_id);` :smile:


resources/assets/js/store/modules/questions.js
```diff
@@ -87,7 +87,7 @@ const mutations = {
 		state.activeFilters = []
 	},
 	[types.QUESTIONS_SET_WITH_ANSWERS] (state, {questions, answers}) {
-		const serialized = state.quiz_questions || {};
+		const serialized = {}
```
adamkarminski: @bognix - I changed it back to `{}`, because using `state.quiz_questions` caused the new questions to just be added at the end of the object AND it caused the list of questions never to refresh. I imagine you wanted to boost performance, but we need to figure out a better way. ;)

bognix: wporzko

PR #555, opened 2017-09-11
Notify only about the first comment;

gitkuba:


app/Listeners/Handlers/CommentPostedHandler.php
```diff
@@ -28,7 +28,11 @@ public function handle(CommentPosted $event, UserNotificationsGate $gate)
 		$excluded = $this->notifyCoCommentators($commentable, $gate, $event);
 
 		$excluded->push($commentableAuthor);
-		$gate->notifyPrivateStream($excluded->pluck('id')->toArray(), $event);
+
+		if ($commentable->comments->count() === 1){
```
adamkarminski: A nie `0`?

PR #579, opened 2017-09-22
Fullscreen

prezestheonlyone:
mobile-friendly fullscreen on Html.vue.


resources/assets/js/components/course/screens/Html.vue
```diff
@@ -1,6 +1,8 @@
 <template>
+
 	<div class="wnl-screen-html" :class="{'wnl-repetitions': isRepetitions}">
-		<div class="content" v-html="content"></div>
+		<div class="content" v-html="content">
```
bognix: `<div class="content v-html="content/>`

PR #630, opened 2017-10-18
Study Buddy self-application check + status fix;

gitkuba:


app/Http/Controllers/Api/PrivateApi/OrdersApiController.php
```diff
@@ -44,20 +45,42 @@ public function putCoupon(UseCoupon $request)
...
 				'errors' => [
-					'code' => [
-						trans('payment.voucher-product-incompatible'),
-					],
+					'code' => $errors,
 				],
 			]);
 		}
 
+		dispatch(new OrderStudyBuddy($order));
```
adamkarminski: Is it supposed to dispatch every time a coupon is added? Seems like an unnecessary step if a regular coupon is used.

PR #725, opened 2017-11-24
Add an option to list all orders with outstanding instalments

adamkarminski:
@gitkuba


gitkuba:
[LGTM](https://www.lgtm.in/i/wqjEdbqPp)

PR #735, opened 2017-12-04
Allow exporting quiz questions by a tag name

adamkarminski:
@gitkuba


adamkarminski:
I thought about changing a tag name to tag id, as it's more reliable... What do you think?

PR #933, opened 2018-06-08
Add a mechanism to timebox news display

adamkarminski:


resources/assets/js/components/course/dashboard/DashboardNews.vue
```diff
@@ -42,23 +44,34 @@
...
+				return CURRENT_NEWS !== ''
+			},
+			hasRequiredRole() {
+				return REQUIRED_ROLE === '' || this.hasRole(REQUIRED_ROLE)
+			},
+			hasSeenNews() {
+				return store.get(this.newsStoreKey)
+			},
+			isNewsTimely() {
+				let now = new Date()
```
bognix: why let not const?

resources/assets/js/components/course/dashboard/DashboardNews.vue
```diff
@@ -42,23 +44,34 @@
...
+				return (DISPLAY_FROM === '' || DISPLAY_FROM < now) &&
```
bognix:
```
return (!DISPLAY_FROM || DISPLAY_FROM < now) && (!DISPLAY_UNTIL || DISPLAY_UNTIL > now)
```
or
```
return (!DISPLAY_FROM instanceof Date || DISPLAY_FROM < now) && (...)
```
the second one is a bit safer in case one day for some reason DISPLAY_FROM will be equal to `"wczoraj"` by accident :)

resources/assets/js/components/course/dashboard/DashboardNews.vue
adamkarminski: ¯\_(ツ)_/¯

PR #935, opened 2018-06-08
fix for no display names in moderators feed

preziu:


resources/assets/js/store/modules/tasks.js
```diff
@@ -128,7 +128,11 @@ function _handleResponse({commit, dispatch}, response, resolve) {
 	const serializedTasks = {}
 	dataArray.forEach(task => {
 		serializedTasks[task.id] = _parseIncludes(included, task)
-		serializedTasks[task.id].assignee = assigneeProfiles[task.assignee_id] || {}
+		serializedTasks[task.id].assignee = Object.values(assigneeProfiles).find(assigneeProfile => {
+			if (assigneeProfile.user_id === task.assignee_id) {
```
bognix: `return assigneeProfile.user_id === task.assignee_id`

PR #959, opened 2018-06-29
Change deletion dependency for Rodo

preziu:


app/Http/Controllers/Api/PrivateApi/UserForgetAccountApiController.php
```diff
@@ -24,7 +25,7 @@ public function patch(Request $request)
 		$password = $request->password;
 
 		if ($user->id == $currentUserId) {
-			if (Hash::check($password, $user->password)) {
+			if (true) {
```
gitkuba: oooo Panie

gitkuba: oooo Panie

PR #965, opened 2018-07-12
Dockerfile

gitkuba:


Dockerfile.nginx
```diff
@@ -0,0 +1,3 @@
+FROM nginx
```
r0bj: I recommend always using image tag to specify image version.

Dockerfile.php
```diff
@@ -0,0 +1,34 @@
...
+# Install JS dependencies & run webpack build 
+#
+
+FROM node:8.11.3-alpine AS js-build
+WORKDIR /src
+
+COPY --from=php-build /src/. .
+RUN yarn run setup
+RUN node node_modules/cross-env/dist/bin/cross-env.js NODE_ENV=production node_modules/webpack/bin/webpack.js --progress --hide-modules --config=webpack.config.js
+RUN rm -rf node_modules && rm -rf resources/assets && rm -f storage/logs/*.log && rm -rf storage/app/public
```
r0bj: It would be probably better to have all those RUN command as one. In current version removing any files does not make any sense because every RUN command is another image layer so those removed files are using space anyway. My proposition:
```
RUN yarn run setup \
  && node node_modules/cross-env/dist/bin/cross-env.js NODE_ENV=production node_modules/webpack/bin/webpack.js --progress --hide-modules --config=webpack.config.js \
  && rm -rf node_modules && rm -rf resources/assets && rm -f storage/logs/*.log && rm -rf storage/app/public
```

PR #1018, opened 2018-08-30
[PLAT-479] Use elasticsearch for users autocomplete

gitkuba:


gitkuba:
TODO: Confirm user profiles are removed from index

PR #1078, opened 2018-10-08
PLAT-611 | Disable instalments for album and remove note about access in invoice for album's instalment

bognix:
When releasing remember to **remove album's instalments row from payment_method_product table**

### DIFF
--- a/lib/Invoice/Invoice.php
+++ b/lib/Invoice/Invoice.php
@@ -587,6 +587,10 @@ class Invoice
 
 	private function getInstalmentsNote($order)
 	{
+		if (empty($order->product->access_start) || empty($order->product->access_end)) {
+			return '';
+		}
+
 		$instalmentsCount = $order->orderInstalments->count();
 		$unpaidInstalments = $order->orderInstalments->where('paid', false);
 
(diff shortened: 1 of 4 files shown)


resources/assets/js/components/user/Order.vue
```diff
@@ -218,7 +218,7 @@
 						<a class=""
 							title="Dodaj lub zmień kod rabatowy"
 							@click="toggleCouponInput"
-							v-if="order.status !== 'closed'">
+							v-if="order.status !== 'clo	sed'">
```
gitkuba: o so chosi?

lib/Invoice/Invoice.php
```diff
@@ -587,6 +587,10 @@ private function price($number)
 
 	private function getInstalmentsNote($order)
 	{
+		if (empty($order->product->access_start) || empty($order->product->access_end)) {
+			return '';
```
gitkuba: Nie wiem co o tym myśleć, może to przegadajmy po prostu.

resources/assets/js/components/user/Order.vue
bognix: ay ay

lib/Invoice/Invoice.php
bognix: discussed offline, couldn't think of anything better for now :(

PR #1089, opened 2018-10-11
[PLAT-616] Render missing invoices

gitkuba:
__Release__
- [ ] Run migration script

### DIFF
--- a/app/Console/Commands/InvoiceRender.php
+++ b/app/Console/Commands/InvoiceRender.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace App\Console\Commands;
+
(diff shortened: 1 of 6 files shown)


app/Console/Commands/InvoiceRender.php
```diff
@@ -0,0 +1,52 @@
...
+     * @var string
+     */
+    protected $signature = 'invoice:render {invoiceId}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Command description';
```
bognix: te cudo możesz usunąć.

PR #1090, opened 2018-10-16
[PLAT-624] Users clean-up

gitkuba:


app/Console/Commands/UserPurge.php
```diff
@@ -0,0 +1,99 @@
...
+				continue;
+			}
+		}
+
+		return;
+	}
+
+	public function purge($user) {
+
+		$user->orders()->delete();
```
kvas-damian: I think we should use correct foreign keys instead of manually remember about all relations.
```
$table->foreign('user_id')
      ->references('id')->on('users')
      ->onDelete('cascade');
```
What if in future we add one more table that has a relation to user? I suspect we forget about adding it here...

app/Console/Commands/UsersCleanUp.php
```diff
@@ -0,0 +1,149 @@
...
+	 * @var string
+	 */
+	protected $description = 'Filter out user accounts';
+
+	/**
+	 * Create a new command instance.
+	 *
+	 * @return void
+	 */
+	public function __construct()
```
kvas-damian: not needed

app/Console/Commands/UsersCleanUp.php
```diff
@@ -0,0 +1,149 @@
...
+
+				case 'skip' :
+					continue;
+					break;
+
+				case 'back' :
+					$i -= 2;
+					break;
+
+				case 'preview changes' :
```
kvas-damian: Should this action skip the current user?

app/Console/Commands/UsersCleanUp.php
```diff
@@ -0,0 +1,149 @@
...
+					$this->showTable($users);
+					break;
+
+				case 'commit changes' :
+					$this->apply($users);
+					break;
+			}
+		}
+
+		return;
```
kvas-damian: when you go through the list of all users and decide what to do the loop will end and there is no chance to commit your changes.  I think it's unexpected

app/Console/Commands/UsersCleanUp.php
```diff
@@ -0,0 +1,149 @@
...
+	 */
+	public function handle()
+	{
+		$roleNames = Role::all()->pluck('name')->toArray();
+		$users = User::with(['userTime'])
+			->where(function ($query) {
+				$query
+					->whereDoesntHave('orders', function ($query) {
+						$query
+							->where('paid', 1);
```
kvas-damian: Is it ok that we try to remove account that could be in the process of payment? Eg. delay from P24?

app/Console/Commands/UserPurge.php
kvas-damian: As an alternative you could hook to `deleting` event: https://laravel.com/docs/5.7/eloquent#events

PR #1116, opened 2018-11-06
PLAT-600 | restore the flashcards state

bognix:


app/Http/Controllers/Api/PrivateApi/UserFlashcardsResultsApiController.php
```diff
@@ -41,4 +41,25 @@ public function post(Request $request) {
...
+	public function fetchMany(Request $request) {
+		$userId = $request->route('userId');
+		$ids = $request->get('flashcards_ids') ?? [];
+		$user = User::fetch($userId);
+
+		if (!Auth::user()->can('view', $user)) {
+			return $this->respondForbidden();
+		}
+
+		$response = UserFlashcardsResults
```
bognix: jak macie pomysł jak to zrobić mądrzej to pomóżcie :)

Potrzebuje wziąć ostatnie odpowiedzi i nie do końca wiedziałam jak to zrobić z poziomu zapytania - stąd `groupBy` na kolekcji

resources/assets/js/components/course/screens/Flashcards.vue
```diff
@@ -176,12 +181,16 @@
 					@media #{$media-query-tablet}
 						flex: 0 0 78px * 3
 
+					&--retake
+						flex: 0 0 78px * 2
```
bognix: 78px do zmiennej

resources/assets/js/store/modules/flashcards.js
```diff
@@ -30,34 +32,37 @@ const mutations = {
...
-	async setFlashcardsSet({commit}, {setId, ...requestParams}) {
+	async setFlashcardsSet({commit, rootGetters}, {setId, ...requestParams}) {
 		try {
 			const {data} = await axios.get(getApiUrl(`flashcards_sets/${setId}`), {
 				params: requestParams
 			})
-			const {included, ...flashcardSet} = data;
+			const {included, ...flashcardSet} = data
+
+			const {data: userResponseData} = await axios.post(getApiUrl(`user_flashcards_results/${rootGetters.currentUserId}`), {
```
bognix: use `current` instead of userId

routes/papi.php
```diff
@@ -278,4 +278,5 @@ function api_action($type, $method)
 	Route::get("{$r['flashcards']}/{id}", 'FlashcardsApiController@get');
 
 	Route::post("{$r['user-flashcards-results']}/{userId}/{flashcardId}", 'UserFlashcardsResultsApiController@post');
+	Route::post("{$r['user-flashcards-results']}/{userId}", 'UserFlashcardsResultsApiController@fetchMany');
```
kvas-damian: I think we need to migrate this route to use currentUserId instead of a parameter

bognix: if you pass `current` instead of the numeric value for ID the user from session will be fetched. There is still check in the Controller to make sure user won't access other users results.

I can remove the /userId from route but not in the same release, because we have 6 webservers and we need to support situation that some webservers have new backend and others have old one I need to leave both routes, and remove in the next release.

app/Http/Controllers/Api/PrivateApi/UserFlashcardsResultsApiController.php
bognix: After my conversation with Adam I need to rewrite it anyway so let's leave it as it is for now.

PR #1118, opened 2018-11-07
PLAT-668 | Allow admins to use all lessons

kvas-damian:


app/Http/Controllers/Api/PrivateApi/UserLessonApiController.php
```diff
@@ -68,8 +68,7 @@ public function putPlan(UpdateLessonsPreset $request, $userId)
 
 		return $this->respondOk([
 			'lessons'        => $controller->transform($lessons),
-			'end_date'       => $plan->last()['start_date']->timestamp,
-			'end_date_human' => $plan->last()['start_date'],
+			'end_date'       => $plan->last()['start_date'] ? $plan->last()['start_date']->timestamp : Carbon::now()->timestamp,
```
bognix: `$plan->last()['start_date']->timestamp ?? Carbon::now()->timestamp`

być może zadziała :)

app/Models/Lesson.php
```diff
@@ -74,6 +82,8 @@ public function startDate($user = null, $editionId = 1)
 
 			if (!is_null($lessonAccess)) {
 				return Carbon::parse($lessonAccess->start_date);
+			} else {
+				return Carbon::now();
```
bognix: return `null` as it was before and update client to handle empty start date.

PR #1123, opened 2018-11-13
PLAT-592 |  flashcard questions editing

kvas-damian:
Added:
* flashcards sets edition on screens editor
* flashcards sets editing
* flashcards questions editing

Notes:
* improved Textarea component to render at least few lines
* Fix `originalData` in Form to not use reference - when editing `data`, `originalData` was also updated by reference so there was no diff.
* Refactored `Autocomplete` component to not be dependent on item rendering component


app/Models/FlashcardsSet.php
```diff
@@ -22,4 +22,15 @@ public function lesson()
 	{
 		return	$this->belongsTo('\App\Models\Lesson');
 	}
+
+	public function syncFlashcards($flashcardIds) {
+		$orderNumber = 0;
+		$flashcardsSync = [];
+		foreach($flashcardIds as $flashcardId) {
```
bognix: how about checking if flashcard with given ID exist?

resources/assets/js/admin/components/flashcards/edit/FlashcardEditor.vue
```diff
@@ -0,0 +1,104 @@
...
+	import { alerts } from 'js/mixins/alerts'
+
+	import Textarea from "js/admin/components/forms/Textarea";
+
+	export default {
+		name: 'FlashcardEditor',
+		components: {
+			'wnl-form-textarea': Textarea,
+		},
+		mixins: [ alerts ],
```
bognix: as an alternative to using this mixin I would reccomend using action `addAutoDismissableAlert`

resources/assets/js/admin/components/flashcards/edit/FlashcardEditor.vue
```diff
@@ -0,0 +1,104 @@
...
+			return {
+				form: new Form({
+					content: null,
+				}),
+				loading: false,
+			}
+		},
+		computed: {
+			flashcardId() {
+				return this.$route.params.flashcardId;
```
bognix: consider passing `flashcardId` as prop instead of computing it from route. This way the form can be reused in different contexts.

resources/assets/js/admin/components/flashcards/FlashcardsSets.vue
```diff
@@ -0,0 +1,37 @@
+<template>
+	<div class="flashcards-sets">
+		<div class="flashcards-container" v-if="isReady">
+			<wnl-flashcards-set-editor v-if="flashcardsSetId" />
```
bognix: I would recommend using child routes here instead of using ifs.

resources/assets/js/admin/components/flashcards/Flashcards.vue
```diff
@@ -0,0 +1,37 @@
+<template>
+	<div class="flashcards">
+		<div class="flashcards-container" v-if="isReady">
+			<wnl-flashcard-editor v-if="flashcardId" />
```
bognix: I would recommend using child routes here instead of using ifs.

resources/assets/js/admin/components/flashcards/Flashcards.vue
```diff
@@ -0,0 +1,37 @@
...
+			<wnl-flashcards-list v-else />
+		</div>
+	</div>
+</template>
+
+<style lang="sass" rel="stylesheet/sass" scoped>
```
bognix: this can be removed.

resources/assets/js/admin/components/flashcards/edit/FlashcardsSetEditor.vue
```diff
@@ -0,0 +1,224 @@
...
+				if (this.form.flashcards.includes(flashcardId)) {
+					this.errorFading('To pytanie jest już w zestawie', 2000)
+				} else {
+					this.form.flashcards.push(flashcardId);
+				}
+				this.flashcard.reset();
+			}
+		},
+		mounted() {
+			this.form.populate(this.flashcardsSetResourceUrl);
```
bognix: if there is no ID - new flashcards set this returns 405 error.

resources/assets/js/admin/components/flashcards/edit/FlashcardsSetEditor.vue
```diff
@@ -0,0 +1,224 @@
...
+			<wnl-form-input
+					name="mind_maps_text"
+					:form="form"
+					v-model="form.mind_maps_text"
+			>
+				Mind mapy
+			</wnl-form-input>
+			<label class="label">Lekcja</label>
+			<span class="select flashcards-set-editor-select">
+				<wnl-select :form="form"
```
bognix: this returns `Duplicate keys detected: 'key'. This may cause an update error`

resources/assets/js/admin/components/flashcards/edit/FlashcardsSetEditor.vue
```diff
@@ -0,0 +1,224 @@
...
+				@input="onDescriptionInput"
+			/>
+			<h4 class="title margin top">Lista pytań</h4>
+			<div class="flashcards-admin">
+				<draggable v-model="form.flashcards" @start="drag=true" @end="drag=false" v-if="areFlashcardsReady">
+					<wnl-flashcards-set-list-item
+							v-for="flashcardId in form.flashcards"
+							:key="flashcardId"
+							:id="flashcardId"
+							:content="allFlashcards.find(flashcard => flashcard.id === flashcardId).content"
```
bognix: `TypeError: Cannot read property 'content' of undefined` w przypadku kiedy ktoś doda ID nieistniejącego flashcarda.

resources/assets/js/admin/components/flashcards/edit/FlashcardsSetEditor.vue
```diff
@@ -0,0 +1,224 @@
...
+		methods: {
+			...mapActions('lessons', { setupLessons: 'setup' }),
+			...mapActions('flashcards', { setupFlashcards: 'setup' }),
+			onDescriptionInput() {
+				this.form.description = this.$refs.descriptionEditor.editor.innerHTML;
+			},
+			removeFlashcard(flashcardId) {
+				this.form.flashcards = this.form.flashcards.filter(id => id !== flashcardId)
+			},
+			flashcardsSetFormSubmit() {
```
bognix: on my local machine submit doesn't work when adding new set. No error is thrown but also no request is issued.

bognix: I don't see how you update the list of flashcards sets once you add new one. The setup method in FlashcardsSets.vue returns the value from store in case it was previously called.

resources/assets/js/admin/store/modules/flashcards.js
```diff
@@ -0,0 +1,58 @@
...
+
+// Initial state
+const state = {
+	ready: false,
+	flashcards: [],
+}
+
+// Getters
+const getters = {
+	isReady: state => state.ready,
```
bognix: you can use `mapState` instead of `mapGetter` to get this value, you wouldn't need this getter if you'll use `mapState`

app/Models/FlashcardsSet.php
kvas-damian: It's an admin tool. Do we really care about such edge-cases for admins?
This row is being ignored when fetching it.

resources/assets/js/admin/components/flashcards/edit/FlashcardsSetEditor.vue
kvas-damian: we're going to migrate to autocomplete solution so the problem should be resolved there

resources/assets/js/admin/components/flashcards/edit/FlashcardsSetEditor.vue
kvas-damian: You're right. I moved the submit button outside of form. Will be fixed.

Also the cache is a  problem.

resources/assets/js/admin/components/lessons/edit/ScreensMetaEditorFlashcards.vue
```diff
@@ -0,0 +1,143 @@
...
+				</ul>
+			</div>
+			<div class="flashcards-set-add">
+				<div class="control">
+					<label class="label">Wybierz zestaw pytań</label>
+					<input class="input" placeholder="Id lub treść aby wyszukać" v-model="flashcardsSetInput"/>
+				</div>
+				<div class="control">
+					<wnl-autocomplete
+							isDown="true"
```
bognix: FYI `isDown` is equal to `"true"` in props (true as String true)

PR #1135, opened 2018-11-20
PLAT-736, PLAT-711 | Flashcards visual improvements

kvas-damian:


resources/assets/js/components/course/screens/flashcards/Flashcards.vue
```diff
@@ -17,26 +17,33 @@
...
-				</div>
-				<div class="flashcards-set__results__single">
-					<span>{{getUnsolvedForSet(set)}}</span>
-					<span>Bez odpowiedzi</span>
-				</div>
-			</div>
-			<div @click="onRetakeSet(set)" class="flashcards-set__retake">
-				<span class="icon"><i class="fa fa-undo"></i></span>
-				ponów cały zestaw
+				<table class="flashcards-set__results__table">
```
bognix: why table?

kvas-damian: The middle column have variable length - depends on the string, that has to be the same in all rows. I'm not sure if we can achieve the same in flexbox

bognix: I see, I in real life and I understand, nevertheless seeing table was unexpected

PR #1136, opened 2018-11-21
PLAT-737 | Make slides matching optional when uploading slides in admin

kvas-damian:


lib/SlideParser/Parser.php
```diff
@@ -90,7 +90,7 @@ public function __construct()
 	 *
 	 * @throws ParseErrorException
 	 */
-	public function parse($fileContents, $screenId = null)
```
bognix: the screen id is used, please check the `SlidesImport` command.

resources/assets/js/plugins/chat-connection.js
```diff
@@ -34,7 +34,7 @@ const WnlSocket = {
     install(Vue, {store}) {
         const onSocketError = (error) => {
             if (error === 'Unauthorized') {
-                window.location.replace('/login');
+                // window.location.replace('/login');
```
bognix: desired change or accidental?

kvas-damian: accidental, thanks for catching

PR #1140, opened 2018-11-22
Don't delete all lessons and roles when deleting a user :)

gitkuba:


app/Console/Commands/UserPurge.php
```diff
@@ -62,22 +62,21 @@ public function handle()
...
-
 		$user->orders()->delete();
 		$user->coupons()->delete();
 		$user->profile()->delete();
 		$user->personalData()->delete();
 		$user->billing()->delete();
 		$user->settings()->delete();
 		$user->userAddress()->delete();
-		$user->roles()->delete();
+		$user->roles()->detach();
```
bognix: what's the guideline here?

On which relations we should use detach and on which delete?

app/Console/Commands/UserPurge.php
```diff
@@ -62,22 +62,21 @@ public function handle()
 	}
 
 	public function purge($user) {
-
 		$user->orders()->delete();
```
kvas-damian: Now I'm not sure if we should be deleteting orders data. Even if there is no user we should have history of all orders. What do you think?

bognix: I agree, given we generate invoices on our side.

app/Console/Commands/UserPurge.php
gitkuba: On many to many or morph type relations we should use `detach()`, every other `delete()`. Not 100% sure though.

app/Console/Commands/UserPurge.php
gitkuba: agree

kvas-damian:
BTW, didn't we agree to move this method to User model hook?

PR #1143, opened 2018-11-26
PLAT-694 | delete quiz questions

kvas-damian:


app/Http/Controllers/Api/PrivateApi/QuizQuestionsApiController.php
```diff
@@ -114,6 +123,38 @@ public function put(UpdateQuizQuestion $request)
...
 	}
 
+	public function delete($id) {
+		$quizQuestion = QuizQuestion::find($id);
+
+		if (!$quizQuestion) {
+			return $this->respondNotFound();
+		}
+
+		if (Auth::user()->can('delete', $quizQuestion)) {
```
bognix: Wouldn't it be better to reuse the "UpdateQuizQuestion" request we already have defined instead of creating new policy?

I feel it's a bit inconsistent across Controllers, sometimes we use middleware, sometimes policy and sometimes request and in all cases we want to check the same thing, if user is an admin and if he can perform this action :)

I would vote for using request or middleware here because it by default returns 403 error.

resources/assets/js/admin/components/quizes/QuizQuestions.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 	<div class="quiz-questions">
-		<form class="quiz-question-search">				
+		<form class="quiz-question-search" v-on:submit.prevent>
```
bognix: @submit.prevent to keep it consistent with remaining usages, I don't think we use v-on directive

resources/assets/js/admin/components/quizes/QuizQuestionEditor.vue
```diff
@@ -250,7 +277,23 @@
 						this.$router.push({name: 'quiz-editor', params: { quizId: data.id }})
 					}, 2000)
 				}
+			},
+			async onDelete() {
+				await this.deleteQuizQuestion(this.questionId);
```
bognix: some error handling please :)

app/Http/Controllers/Api/PrivateApi/QuizQuestionsApiController.php
kvas-damian: We have a middleware already in place, so I removed this check

PR #1146, opened 2018-11-26
PLAT-744 | Invoice data

bognix:


resources/assets/js/admin/components/users/UserBilling.vue
```diff
@@ -0,0 +1,49 @@
+<template lang="html">
```
kvas-damian: Maybe we could reuse code from https://github.com/bethinkpl/wnl-platform/blob/b2b652e467443ed918f3d3f66427ce81d62248cb/resources/assets/js/components/user/MyBillingData.vue ?

PR #1147, opened 2018-11-26
PLAT-745 | Display User Lessons

bognix:


resources/assets/js/admin/components/users/UserPlan.vue
```diff
@@ -1,13 +1,85 @@
...
+				</div>
+			</div>
+		</div>
+		<template v-if="userLessons.length">
+			<input v-model="filterPhrase" class="input margin bottom" placeholder="filtruj..." ref="filterInput">
+			<table class="table is-bordered">
+				<thead>
+					<tr>
+						<th>ID lekcji</th>
+						<th>Lekcja</th>
```
kvas-damian: Maybe we could also show a group name the lesson belongs to?

bognix: makes sense :) I'll update.

PR #1151, opened 2018-11-29
PLAT-746 | Admin dashboardNews preview

kvas-damian:


resources/assets/js/components/global/form/Form.vue
```diff
@@ -155,6 +155,10 @@
 
 					dataModel[name] = defaultValue
 					defaults[name] = defaultValue
+
+					child.$on('input', (value) => {
```
bognix: can you declare @input on <slot> to avoid this usage of children? It's really ugly.

resources/assets/js/components/course/dashboard/DashboardNewsContent.vue
```diff
@@ -0,0 +1,26 @@
+<template>
+	<div class="notification content">
+		<button class="delete" @click="$emit('onClose')"></button>
+		<p class="has-text-centered"><strong v-html="slug"></strong></p>
+		<span v-html="message"></span>
+	</div>
+</template>
+
+<script>
+	import store from 'js/services/messagesStore'
```
bognix: I think neither store nor getters are used here.

PR #1152, opened 2018-11-29
PLAT-764 | add user lesson

bognix:


resources/assets/js/admin/components/users/UserPlan.vue
```diff
@@ -30,17 +30,86 @@
...
+			<div class="level wnl-screen-title">
+				<div class="level-left">
+					<div class="level-item big strong">
+						Dodaj Lekcje
+					</div>
+				</div>
+			</div>
+			<div class="control">
+				<label class="label">Wyszukaj Lekcję</label>
+				<input class="input" placeholder="nazwa aby wyszukać" v-model="lessonInput"/>
```
kvas-damian: I think the placeholder is not grammatically correct. Maybe "Wpisz nazwę aby wyszukać"?

resources/assets/js/admin/components/users/UserPlan.vue
```diff
@@ -51,23 +120,93 @@
...
 
 				return this.userLessons.filter(({lesson}) => lesson.toLowerCase().startsWith(this.filterPhrase.toLowerCase()));
+			},
+			autocompleteLessonsItems() {
+				if (this.lessonInput === '') {
+					return [];
+				}
+
+				return this.lessons
+					.filter(lesson => !this.userLessons.find(({lesson_id}) => lesson_id === lesson.id) &&
```
kvas-damian: Don't we need to exclude selected lessons as well?

bognix: we can actually but they won't be inserted any way because I'm checking inside addLesson method if lesson is already selected.

PR #1158, opened 2018-12-03
Admins and moderators subscription is always active

rogatty:
Admins and moderators subscription is always active, even if there are no min or max dates set.


kvas-damian:
Tested, works! :tada:

PR #1165, opened 2018-12-06
PLAT-794  | Allow {{currentUserName}} in dashboard news message

rogatty:
https://bethink.atlassian.net/browse/PLAT-794


resources/assets/js/admin/components/dashboardNews/DashboardNewsEdit.vue
```diff
@@ -53,15 +70,21 @@
...
+				return this.isEdit ? 'put' : 'post';
 			},
 			isEdit() {
 				return this.id !== 'new';
-			}
+			},
+			messageArguments() {
+				return {
+					currentUserName: this.currentUserName
+				};
```
rogatty: We'll need to maintain this list of arguments in two places. Would it be better to use a mixin?

resources/assets/js/admin/components/dashboardNews/DashboardNewsEdit.vue
```diff
@@ -9,26 +9,43 @@
 				@submitSuccess="onSubmitSucess"
 				@change="onChange"
 		>
-			<wnl-text name="slug">Slug</wnl-text>
+			<wnl-text name="slug">Tytuł</wnl-text>
 			<wnl-textarea name="message">Treść</wnl-textarea>
+			<p>Możesz użyć następujących parametrów:</p>
+			<ul class="message-help">
+				<li v-for="argumentKey in Object.keys(messageArguments)">
```
kvas-damian: Please set `:key` attribute to make Vue life easier.

bognix: v-for="(value, key) in messageArguments"

resources/assets/js/admin/components/dashboardNews/DashboardNewsEdit.vue
```diff
@@ -9,26 +9,43 @@
...
 </template>
 
 <style lang="sass" scoped>
 	.notification
 		max-width: 900px
+
+	.message-help
+		list-style: disc
+		margin-bottom: 10px
+		/deep/ li
```
kvas-damian: maybe it's better to style based on classes?

resources/assets/js/admin/components/dashboardNews/DashboardNewsEdit.vue
rogatty: Based on https://vuejs.org/v2/guide/list.html#key, I think it makes sense only when iterating over an array of objects.

resources/assets/js/admin/components/dashboardNews/DashboardNewsEdit.vue
rogatty: @bognix `named exports are not supported in *.vue files.`

PR #1166, opened 2018-12-06
PLAT-796 | Track dashboard news events

rogatty:
https://bethink.atlassian.net/browse/PLAT-796

- Use https://dennisreimann.de/articles/delegating-html-links-to-vue-router.html to open relative URLs without reloading the page
- Track open, close and click_link in dashboard news
- Fix events which were using `actions` instead of `action`

### DIFF
--- a/resources/assets/js/components/course/dashboard/DashboardNews.vue
+++ b/resources/assets/js/components/course/dashboard/DashboardNews.vue
@@ -4,13 +4,14 @@
 			:message="dashboardNews.message"
 			:messageArguments="messageArguments"
 			:slug="dashboardNews.slug"
-			@onClose="seenCurrentNews"
+			@onClose="closed"
+			@onContentClick="contentClicked"
 	/>
 </template>
 
 <script>
-	import store from 'js/services/messagesStore';
 	import {mapGetters} from 'vuex';
+	import store from 'js/services/messagesStore';
 	import WnlDashboardNewsContent from 'js/components/course/dashboard/DashboardNewsContent';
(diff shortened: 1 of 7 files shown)


resources/assets/js/components/course/dashboard/DashboardNews.vue
```diff
@@ -36,13 +37,62 @@
 			},
 		},
 		methods: {
-			seenCurrentNews() {
+			closed() {
+				this.track(context.dashboard.features.news_message.actions.close.value);
```
bognix: since this component is only related with dashboard news maybe you could define in data:
`feature: context.dashboard.features.news_message` and later use `this.feature....`
so you don't have to repeat this long chain.

resources/assets/js/components/course/dashboard/DashboardNews.vue
```diff
@@ -36,13 +38,62 @@
...
 				this.showNews = false;
 				store.set(this.newsStoreKey, true);
 			},
+			contentClicked({target}) {
+				const href = target.getAttribute('href');
+
+				if (target && href) {
+					this.track(this.featureContext.actions.click_link.value);
+
+					if (/^https?:\/\//.test(href)) {
```
kvas-damian: What if admin uses a link to our platform including proto and domain?
Can we add support for proto-relative links?

rogatty: Let's add info about it (https://github.com/bethinkpl/wnl-platform/pull/1166/commits/26572f66e7602ea95651501c252893810bb9258e) and implement parsing full URLs only if people complain about it.

PR #1169, opened 2018-12-10
PLAT-810 | Add description of available message arguments

rogatty:
https://bethink.atlassian.net/browse/PLAT-810


resources/assets/js/admin/components/dashboardNews/DashboardNewsEdit.vue
```diff
@@ -23,7 +23,7 @@
 						class="message-argument"
 						v-for="(value, key) in messageArguments" :key="key"
```
kvas-damian: Maybe we should iterate over `messageArgumentsDescription`? Currently it's possible these variables get out of sync.

rogatty: @kvas-damian how about https://github.com/bethinkpl/wnl-platform/pull/1169/commits/a0a6939718995c99c6fc4679deb0bc058dc43be9?

kvas-damian:
Can you investigate why tests are failing on travis?

PR #1172, opened 2018-12-11
PLAT-789 | Manage groups and lessons

kvas-damian:


app/Http/Controllers/Api/PrivateApi/CoursesApiController.php
```diff
@@ -10,4 +12,23 @@ public function __construct(Request $request)
...
+
+		if (empty($course)) {
+			return $this->respondNotFound();
+		}
+
+		$course->update($request->all());
+
+		foreach($course->groups as $group) {
+			$group->order_number = array_search($group->id, $request->groups) + 1;
+			$group->save();
```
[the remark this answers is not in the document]
rogatty: https://stackoverflow.com/questions/15105640/laravel-eloquent-orm-transactions

app/Http/Controllers/Api/Transformers/GroupTransformer.php
```diff
@@ -13,25 +13,32 @@ class GroupTransformer extends ApiTransformer
...
+		$this->parent = $parent;
 	}
 
 	public function transform(Group $group)
 	{
-		return [
+		$data = [
 			'id'            => $group->id,
 			'name'          => $group->name,
 			'editions'      => $group->course_id
```
rogatty: Here we have `editions` and in `EditionTransformer.php` we use `new GroupTransformer(['editionId' => $edition->id])`. We'll return:
```
[
  ...,
  'editions' => 'x',
  'editionId' => 'y'
]
```
Is this correct?

app/Http/Requests/Course/UpdateCourse.php
```diff
@@ -0,0 +1,30 @@
...
+class UpdateCourse extends FormRequest
+{
+	/**
+	 * Determine if the user is authorized to make this request.
+	 *
+	 * @return bool
+	 */
+	public function authorize()
+	{
+		return $this->user()->isAdmin();
```
rogatty: It's behind the admin middleware. Do we need to check it twice?

app/Http/Controllers/Api/PrivateApi/CoursesApiController.php
```diff
@@ -10,4 +12,23 @@ public function __construct(Request $request)
...
+		$course = Course::find($id);
+
+		if (empty($course)) {
+			return $this->respondNotFound();
+		}
+
+		$course->update($request->all());
+
+		foreach($course->groups as $group) {
+			$group->order_number = array_search($group->id, $request->groups) + 1;
```
bognix: `foreach($course->groups as $index => $group)`
and then use `$index + 1`.

I don't think the array_search is the most performant solution.

app/Http/Controllers/Api/Transformers/GroupTransformer.php
bognix: the naming is very misleading in Fractal :(

app/Http/Requests/Course/UpdateCourse.php
```diff
@@ -0,0 +1,30 @@
...
+
+	/**
+	 * Get the validation rules that apply to the request.
+	 *
+	 * @return array
+	 */
+	public function rules()
+	{
+		return [
+			'name'   => 'required|string',
```
rogatty: How about adding `groups` as a required field of array type?

app/Http/Controllers/Api/Transformers/GroupTransformer.php
```diff
@@ -13,25 +13,32 @@ class GroupTransformer extends ApiTransformer
...
+
+		return $data;
 	}
 
 	public function includeLessons(Group $group)
 	{
 		$lessons = $group->lessons->sortBy('order_number');
 		$meta = collect([
-			'editionId' => $this->editionId,
+			'editionId' => $this->parent['editionId'] ?? null,
```
bognix: I'm not sure if the editionId is used anywhere on lesson. It's an odd situation that we pass info about "grandparent" (edition) to the lesson. There is no other connection between those two resources, only through groups.

app/Http/Requests/Course/UpdateLesson.php
```diff
@@ -24,8 +24,9 @@ public function authorize()
 	public function rules()
 	{
 		return [
-			'name'   => 'string',
-			'groups' => 'numeric',
+			'name' => 'required|string',
+			'group_id' => 'required|numeric',
```
bognix: maybe something like this makes sense here: https://laravel.com/docs/5.7/validation#rule-exists

resources/assets/js/admin/components/courses/CourseEditor.vue
```diff
@@ -0,0 +1,80 @@
...
+			},
+			resourceRoute() {
+				return this.isEdit ? `courses/${this.id}?include=groups` : 'courses?include=groups';
+			},
+		},
+		components: {
+			WnlFormText,
+			WnlForm,
+			WnlSubmit,
+            WnlSortable,
```
bognix: dziwne wciecia

app/Http/Requests/Course/UpdateGroup.php
```diff
@@ -0,0 +1,30 @@
...
+
+	/**
+	 * Get the validation rules that apply to the request.
+	 *
+	 * @return array
+	 */
+	public function rules()
+	{
+		return [
+			'name'   => 'required|string',
```
rogatty: How about adding `lessons` as a required field of array type?

app/Http/Requests/Course/UpdateLesson.php
rogatty: I think it would be:
```suggestion
			'group_id' => 'required|numeric|exists:groups,id',
```

database/migrations/2018_12_10_114038_groups__drop_required_role.php
```diff
@@ -0,0 +1,32 @@
...
+use Illuminate\Database\Migrations\Migration;
+
+class GroupsDropRequiredRole extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
```
rogatty: Indentation is mixed up in this file.

app/Http/Controllers/Api/PrivateApi/CoursesApiController.php
kvas-damian: I need to use index based on the order provided by user in the request, that's why I need to merge groups from database with the order in the request.

app/Http/Controllers/Api/Transformers/GroupTransformer.php
kvas-damian: I wasn't sure about it. I'll check and remove it if not required.

app/Http/Requests/Course/UpdateCourse.php
kvas-damian: I'm ok with having it in two places. IMO it's better to protect two time than miss the authorization when someone moves the route.
Is it ok for you?

app/Http/Requests/Course/UpdateCourse.php
kvas-damian: Done

app/Http/Controllers/Api/Transformers/GroupTransformer.php
kvas-damian: Removed

app/Http/Requests/Course/UpdateGroup.php
kvas-damian: Done

resources/assets/js/admin/components/lessons/edit/LessonEditor.vue
```diff
@@ -104,35 +111,40 @@
...
-				this.form.group_id = this.form.groups
-				this.form.put(this.resourceUrl)
+				this.loading = true;
+				this.form[this.method](this.resourceUrl)
 					.then(response => {
-						this.loading = false
-						this.successFading('Lekcja zapisana!', 2000)
-						this.form.originalData = this.form.data()
+						this.loading = false;
+						this.successFading('Lekcja zapisana!', 2000);
```
bognix: consider replacing `successFading` and `errorFading` with `addAutoDismissableAlert`. Let's reduce the number of places we use this mixin in, let it die!

app/Http/Controllers/Api/Transformers/GroupTransformer.php
```diff
@@ -11,27 +11,30 @@ class GroupTransformer extends ApiTransformer
...
-		return [
+		$data = [
 			'id'            => $group->id,
 			'name'          => $group->name,
-			'editions'      => $group->course_id
+			'courses'      => $group->course_id
 		];
+
+		if ($this->parent) {
+			$data = array_merge($data, $this->parent);
```
rogatty: In `LessonTransformer.php` we have `$data = array_merge($data, $this->parent->toArray());`
How do you feel about keeping the types consistent?

app/Http/Controllers/Api/Transformers/TagsTaxonomyTransformer.php
```diff
@@ -9,20 +9,11 @@
...
 	{
 		return [
-			'id'   => $edition->id,
-			'name' => $edition->name,
+			'id'   => $tagsTaxonomy->id,
+			'name' => $tagsTaxonomy->name,
 		];
 	}
-
-	public function includeGroups(Edition $edition)
```
rogatty: Am I correct that this can be removed because we don't have `$availableIncludes = ['groups']` in this file, so it didn't work anyway?

resources/assets/js/admin/components/groups/GroupsListItem.vue
```diff
@@ -0,0 +1,22 @@
+<template>
+	<li class="margin bottom">
+		<router-link :to="to">{{id}}. {{name}}</router-link>
```
rogatty: `{{id}}. {{name}}` might be misleading. I think most people would assume this format means order position.

resources/assets/js/mixins/form-input.js
```diff
@@ -17,6 +17,9 @@ export var formInput = {
 		getErrors() {
 			return this.getterFunction('getErrors', this.name);
 		},
+		data() {
+			return this.getter('getData');
```
rogatty: In `Form.vue` this computed property is called `formData()`. What do you think about keeping it consistent?

rogatty: This one also returns the whole form data, right?

rogatty:
We updated some tests but didn't add any new test cases. How about testing the new `post` and `put` actions?

app/Http/Controllers/Api/Transformers/TagsTaxonomyTransformer.php
kvas-damian: yes

resources/assets/js/admin/components/groups/GroupsListItem.vue
kvas-damian: We use this pattern in multiple places and noone complained about it. Let's get back to it when we get that feedback.

app/Http/Controllers/Api/Transformers/GroupTransformer.php
kvas-damian: Actually, it's more common to use an array as type of `$parent` - not `Collection` as in `LessonTransformer`. We can unify that, but let's do that separately - there're already too many changes in this PR.

kvas-damian: https://github.com/bethinkpl/wnl-platform/pull/1181

PR #1173, opened 2018-12-11
Introduce eslint and run it on Travis

rogatty:
I enabled only the most basic rules. Most of the changes are missing semicolons.


bognix:
@rogatty the checks are failing because some dependencies cannot be installed. Could you specify the node version in package.json?

rogatty:
@bognix unfortunately it's more complicated than that. Hopefully the method from https://austinpray.com/ops/2015/09/20/change-travis-node-version.html will fix it.

PR #1174, opened 2018-12-11
PLAT-818 | Store progress when weird things in Redis

bognix:


app/Http/Controllers/Api/PrivateApi/UserStateApiController.php
```diff
@@ -77,6 +77,12 @@ public function putLesson(Request $request, $id, $courseId, $lessonId)
 	{
 		$lesson = $request->lesson;
 
+		if (empty($lesson)) {
+			\Log::error('Trying to save empty progress for lesson');
+			// return OK because we want to make it transparent for a user.
+			return $this->respondOk();
```
bognix: I want to gather more data about such cases. In Sentry we have an access to request object so maybe we can find out more with it.

app/Console/Commands/StoreProgress.php
```diff
@@ -103,10 +103,11 @@ private function storeProgress($key, $userId, $bar)
 					$lessonKey = UserStateApiController::getLessonRedisKey($userId, 1, $lessonId);
 					$lessonProgressRaw = $this->redis->get($lessonKey);
 
-					if (!$lessonProgressRaw) continue;
-
+					if (empty($lessonProgressRaw)) continue;
 					$lessonProgress = json_decode($lessonProgressRaw);
 
+					if (empty($lessonProgress)) continue;
```
kvas-damian: Do we want to silently exit or maybe we could log something?

rogatty: I don't think there is any value in logging it here as there is no context.
https://github.com/bethinkpl/wnl-platform/pull/1174/files#diff-9b6949db0a4ab114e0554d7930142209R81 should be enough.

PR #1180, opened 2018-12-12
Enable eslint-plugin-vue and use autofix for indentation and semicolons

rogatty:
After enabling `plugin:vue/recommended` and running `yarn run lint:fix` there are `758 problems (333 errors, 425 warnings)` to fix manually, so I'll leave it for later.


rogatty:
@kvas-damian I'll need to fix the conflicts anyway and the longer we wait the more to fix.

PR #1184, opened 2018-12-13
PLAT-770 | presentables order numbers integrity check

bognix:


tests/DataIntegrity/PaymentsCheck.php
```diff
@@ -14,9 +14,9 @@ private function checkInstalments() {
 			->havingRaw('SUM(order_instalments.paid_amount) != orders.paid_amount')
 			->get();
 
-		if (!empty($incorrectInstalments)) {
+		if ($incorrectInstalments->count() > 0) {
 			$this->handleError(__METHOD__, [
-				'orders.id' => $incorrectInstalments->pluck('id'),
+				'orders.id' => $incorrectInstalments->pluck('id')->toArray(),
```
rogatty: Shouldn't this be `->all()`? https://laravel.com/docs/5.7/collections#method-pluck

bognix: toArray is safer. If the values would be eloquent models `all` would change the collection to array of eloquent models.

PR #1187, opened 2018-12-16
Install laravel-ide-helper and generate docs

rogatty:
https://github.com/barryvdh/laravel-ide-helper

### DIFF
--- a/app/Models/Structure.php
+++ b/app/Models/Structure.php
@@ -11,10 +11,6 @@ class Structure extends Model
 
     protected $fillable = ['name', 'subject_id'];
 
-	public function subject() {
-		return $this->belongsTo('\App\Models\Subject');
-	}
-
 	public function screens() {
 		return $this->belongsToMany('\App\Models\Screen', 'screens');
 	}
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -1,0 +1,183 @@
+<?php
+
+return array(
+
+    /*
+    |--------------------------------------------------------------------------
+    | Filename & Format
+    |--------------------------------------------------------------------------
+    |
+    | The default filename (without extension) and the format (php or json)
+    |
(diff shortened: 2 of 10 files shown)


config/ide-helper.php
```diff
@@ -0,0 +1,183 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | The default filename (without extension) and the format (php or json)
+    |
+    */
+
+    'filename'  => '_ide_helper',
+    'format'    => 'php',
+    
+    'meta_filename' => '.phpstorm.meta.php',
```
kvas-damian: Are we going to use this meta file? I don't see it in the PR.

rogatty: This file was generated after running:
```
php artisan vendor:publish --provider="Barryvdh\LaravelIdeHelper\IdeHelperServiceProvider" --tag=config
```
I did it only to set https://github.com/bethinkpl/wnl-platform/blob/dc38c1993be041ad66274270632db3a50694f188/config/ide-helper.php#L137-L141

I don't think we need the meta file for now, but if we ever do and want to run `ide-helper:meta` command, then this is the default path it will use. I'd leave it as is.

app/Models/Structure.php
```diff
@@ -11,10 +11,6 @@ class Structure extends Model
 
     protected $fillable = ['name', 'subject_id'];
 
-	public function subject() {
-		return $this->belongsTo('\App\Models\Subject');
```
rogatty: It was removed in https://github.com/bethinkpl/wnl-platform/commit/2c7681e6cc90fdbb74f90205604a785677923ed4#diff-f2f67bffd4b2ebfc7d59722666ced328L1
Here is the ticket for the further clean up: https://bethink.atlassian.net/browse/PLAT-840.

PR #1191, opened 2018-12-17
fix wrong column name in CommentsApiController::query

bognix:


app/Http/Controllers/Api/PrivateApi/CommentsApiController.php
```diff
@@ -80,8 +80,8 @@ public function query(Request $request) {
 			$commentsQuery->where('commentable_type', $commentableType);
 		}
 
-		if ($request->has('comment_id')) {
-			$commentsQuery->where('comment_id', $request->get('comment_id'));
+		if ($request->has('id')) {
```
rogatty: Are you sure about this one? If there was an error about `comment_id` then the request probably uses `comment_id`

rogatty: https://github.com/bethinkpl/wnl-platform/blob/master/resources/assets/js/store/modules/comments.js#L25

kvas-damian: frontend still uses `comment_id`

PR #1192, opened 2018-12-17
PLAT-828 | CRUD for tags + batch move of taggables

rogatty:
https://bethink.atlassian.net/browse/PLAT-828


app/Http/Controllers/Api/PrivateApi/TagsApiController.php
```diff
@@ -28,4 +28,72 @@ public function byName(Request $request) {
...
+
+		if ($request->get('name') !== $tag->name && $tag->isCategoryTag()) {
+			return $this->respondUnprocessableEntity([
+				'message' => 'Na podstawie nazwy tego taga tworzymy slajdy dla kategorii. Nie możesz zmienić mu nazwy.'
+			]);
+		}
+
+		$tag->update($request->all());
+
+		return $this->respondOk();
```
kvas-damian: We could return updated object as a response.

app/Http/Controllers/Api/PrivateApi/TagsApiController.php
```diff
@@ -28,4 +28,72 @@ public function byName(Request $request) {
...
+				'message' => 'Ten tag jest częścią taksonomii. Nie możesz go usunąć.'
+			]);
+		}
+
+		$taggables = Taggable::where('tag_id', $tag->id)
+			->get();
+
+		\DB::transaction(function () use ($tag, $taggables) {
+			foreach ($taggables as $taggable) {
+				$taggable->delete();
```
kvas-damian: You could use eloquent hooks to make sure this operation is executed on every delete, not just in this controller: https://laravel.com/docs/5.7/eloquent#events

app/Http/Controllers/Api/Transformers/TagTransformer.php
```diff
@@ -1,27 +1,26 @@
...
+	public function transform(Tag $tag) {
 		$data = [
-			'id'   => $tag->id,
+			'id' => $tag->id,
 			'name' => $tag->name,
+			'description' => $tag->description,
+			'color' => $tag->color,
+			'is_taggable' => $tag->isTaggable(),
+			'is_rename_allowed' => $tag->isRenameAllowed(),
+			'is_delete_allowed' => $tag->isDeleteAllowed()
```
kvas-damian: I'm not sure if we should put these fields in the transformer. It's used in multiple places on the platform and it adds unnecessary DB queries.

app/Http/Requests/Course/UpdateTag.php
```diff
@@ -0,0 +1,29 @@
...
+	/**
+	 * Get the validation rules that apply to the request.
+	 *
+	 * @return array
+	 */
+	public function rules() {
+		return [
+			'name' => 'required|string|max:255|unique:tags',
+			'description' => 'string|max:1000',
+			'color' => 'string|max:6'
```
kvas-damian: I'd suggest validation for hex color values. Otherwise someone will put `red` or something similar here.

app/Models/Tag.php
```diff
@@ -3,20 +3,56 @@
...
+
+	public function isTaggable() {
+		return DB::table('taggables')
+			->select('tag_id')
+			->where('tag_id', $this->id)
+			->exists();
+	}
+
+	public function isProtectedTaggable() {
+		return DB::table('taggables')
```
kvas-damian: I think it's better to use Eloquent relation instead of performing a separate query.

app/Http/Controllers/Api/PrivateApi/TagsApiController.php
rogatty: @bognix @gitkuba is there a reason we're not using foreign keys? Seems like a task for DB, not for PHP hooks.

app/Http/Controllers/Api/PrivateApi/TagsApiController.php
rogatty: I checked other PUT resources and we don't have a consistent convention. I'd say return it only if it's needed on the front end.

app/Models/Tag.php
rogatty: Can I use an existing relation or do I need to add a new one? This is required only in the admin panel and I don't want to add unnecessary queries for regular users. I'd need some help with it.

app/Http/Controllers/Api/PrivateApi/TagsApiController.php
kvas-damian: We had that discussion recently and we decided to use Laravel hooks instead of foreign keys. We have multiple morph relations which aren't easy to handle by foreign key.
In future we can consider changing that decision.

rogatty: @kvas-damian if we use events then we can't use a transaction. It means that we either have to delete the taggables first or have a broken state until all deletions are finished. If we have a taggable related to a deleted tag, then most likely something will break.

kvas-damian: Observers aren't executed asynchronously, so it should be possible to catch them in a transaction. Do you have a specific example?

kvas-damian: Also you can override `delete` method on the model.

app/Http/Controllers/Api/PrivateApi/TagsApiController.php
```diff
@@ -1,13 +1,13 @@
...
+use App\Http\Requests\Course\UpdateTag;
 use App\Models\Tag;
+use App\Models\Taggable;
 use Illuminate\Http\Request;
 
-class TagsApiController extends ApiController
-{
-	public function __construct(Request $request)
-	{
+class TagsApiController extends ApiController {
```
kvas-damian: https://www.php-fig.org/psr/psr-2/

resources/assets/js/admin/components/tags/TagDelete.vue
```diff
@@ -0,0 +1,114 @@
+<template>
+	<div>
+		<button
+			class="button is-danger margin right"
+			type="button"
+			@click="onClick"
+			:disabled="!isDeleteAllowed"
```
kvas-damian: I think we could add a info for admin saying why the tag cannot be deleted.

resources/assets/js/admin/components/tags/TaggablesMover.vue
```diff
@@ -0,0 +1,53 @@
...
+import {mapActions} from 'vuex';
+import {Form as WnlForm, Text as WnlFormText, Submit as WnlSubmit} from 'js/components/global/form';
+
+export default {
+	name: 'TaggablesMover',
+	props: {
+		beforeSubmit: {
+			default: null,
+		},
+		sourceTagId: {},
```
kvas-damian: please add type and default/required

resources/assets/js/components/global/form/Text.vue
```diff
@@ -36,6 +37,9 @@ export default {
 		},
 		placeholder: {
 			type: String
+		},
+		disabled: {
+			type: Boolean
```
kvas-damian: default value

app/Http/Controllers/Api/PrivateApi/TaggablesApiController.php
```diff
@@ -0,0 +1,35 @@
...
+
+		if (!$sourceTag) {
+			return $this->respondNotFound('Source tag does not exist');
+		}
+
+		$targetTag = Tag::find($request->get('target_tag_id'));
+
+		$taggablesMoved = Taggable::where('tag_id', $sourceTag->id)
+			// TODO Is this necessary? If so, replace this comment with the reason for it
+			->whereNotIn('taggable_type', Taggable::PROTECTED_TAGGABLE_TYPES)
```
kvas-damian: Instead of this where I'd suggest raising an error. 
Or we update everything or nothing.

resources/assets/js/admin/components/tags/TagDelete.vue
```diff
@@ -0,0 +1,103 @@
...
+import WnlTaggablesMover from 'js/admin/components/tags/TaggablesMover';
+
+export default {
+	name: 'TagDelete',
+	components: {
+		'wnl-modal': Modal,
+		'wnl-taggables-mover': WnlTaggablesMover,
+	},
+	props: {
+		id: {},
```
kvas-damian: Type, required/default

app/Http/Controllers/Api/PrivateApi/TaggablesApiController.php
bognix: agree

app/Http/Controllers/Api/PrivateApi/TagsApiController.php
```diff
@@ -28,4 +28,67 @@ public function byName(Request $request) {
...
+
+		if (!$tag) {
+			return $this->respondNotFound();
+		}
+
+		if ($request->get('name') !== $tag->name && $tag->isCategoryTag()) {
+			return $this->respondUnprocessableEntity([
+				'errors' => [
+					'name' => [
+						'Na podstawie nazwy tego taga tworzymy slajdy dla kategorii. Nie możesz zmienić mu nazwy.'
```
bognix: In API please use translations.

app/Models/Tag.php
```diff
@@ -3,28 +3,62 @@
...
 	{
 		return $this->hasMany('App\Models\Taggable');
 	}
+
+	public function taxonomies()
+	{
+		return $this->belongsToMany('App\Models\Taxonomy', 'tags_taxonomy');
+	}
+
+	public function delete() {
```
bognix: you are overriding the Eloquent delete model here. Maybe instead you can use observable and delete taggables in observable?

app/Models/Tag.php
```diff
@@ -3,28 +3,62 @@
...
+		\DB::transaction(function () {
+			$this->taggables()->delete();
+			parent::delete();
+		});
+	}
+
+	public function isInTaxonomy() {
+		return $this->taxonomies()->exists();
```
bognix: $this->taxonomies() returns query builder but $this->taxonomies returns collection. You can use collection and check if it's empty instead of calling `exists` which probably does some query under the hood.

app/Models/Tag.php
```diff
@@ -3,28 +3,62 @@
...
+	}
+
+	public function isCategoryTag() {
+		return Category::where('name', $this->name)->exists();
```
bognix: same as above.

app/Http/Controllers/Api/PrivateApi/TagsApiController.php
rogatty: Do you mean "Code MUST use 4 spaces for indenting, not tabs." 😛

resources/assets/js/admin/components/tags/TagDelete.vue
```diff
@@ -0,0 +1,113 @@
...
+			class="button is-danger margin right"
+			type="button"
+			@click="onClick"
+			:disabled="!isDeleteAllowed"
+			:title="!isDeleteAllowed && 'Nie można usunąć tego tagu ponieważ należy do taksonomii lub jest kategorią lub taguje lekcję albo stronę'"
+		>Usuń</button>
+		<wnl-modal
+			:isModalVisible="isTaggablesMoverVisible"
+			@closeModal="onCloseModal"
+			v-if="isTaggablesMoverVisible"
```
bognix: I know this is how modals are implemented but it looks a bit odd that you need to pass to modal if it's visible + add v-if here to make sure it's not rendered. I think the v-if could be moved to the internals of wnl-modal

resources/assets/js/admin/components/tags/TagDelete.vue
```diff
@@ -0,0 +1,113 @@
...
+import {mapActions} from 'vuex';
+import {getApiUrl} from 'js/utils/env';
+import Modal from 'js/components/global/Modal';
+import WnlTaggablesMover from 'js/admin/components/tags/TaggablesMover';
+
+export default {
+	name: 'TagDelete',
+	components: {
+		'wnl-modal': Modal,
+		'wnl-taggables-mover': WnlTaggablesMover,
```
bognix: WnlTaggablesMover is enough, you don't need to specify the key.

resources/assets/js/admin/components/tags/TagDelete.vue
```diff
@@ -0,0 +1,113 @@
...
+			</div>
+		</wnl-modal>
+	</div>
+</template>
+
+<script>
+import axios from 'axios';
+import {mapActions} from 'vuex';
+import {getApiUrl} from 'js/utils/env';
+import Modal from 'js/components/global/Modal';
```
bognix: you can import it as WnlModal, then the 'wnl-modal' key in `components` object is redundant.

resources/assets/js/admin/components/tags/TagDelete.vue
```diff
@@ -0,0 +1,113 @@
...
+		isDeleteAllowed: {
+			type: Boolean,
+			default: false,
+		},
+		taggablesCount: {
+			type: Number,
+			default: 0,
+		}
+	},
+	data: () => {
```
bognix:
```
data() {
}
```

so `this` works fine inside this method.

resources/assets/js/admin/components/tags/TagDelete.vue
```diff
@@ -0,0 +1,113 @@
...
+		return {
+			isTaggablesMoverVisible: false,
+		};
+	},
+	computed: {},
```
bognix: redundant

resources/assets/js/admin/components/tags/TagDelete.vue
```diff
@@ -0,0 +1,113 @@
...
+		async deleteTag() {
+			try {
+				await axios.delete(getApiUrl(`tags/${this.id}`));
+				this.addAutoDismissableAlert({
+					text: 'Tag został usunięty',
+					type: 'success',
+				});
+
+				this.$emit('tagDeleted');
+			} catch ({response: {data: {message}}}) {
```
bognix: that's really brave assumption to claim that data and message will be always inside response :)

resources/assets/js/admin/components/tags/TagDelete.vue
```diff
@@ -0,0 +1,113 @@
...
+				this.deleteTag();
+			}
+		},
+		showTaggablesMover() {
+			this.isTaggablesMoverVisible = true;
+		},
+		async onClick() {
+			if (this.taggablesCount > 0) {
+				this.showTaggablesMover();
+			} else {
```
bognix: `} else if (...) {`

resources/assets/js/admin/components/tags/TagEditor.vue
```diff
@@ -0,0 +1,120 @@
...
+import WnlTagDelete from 'js/admin/components/tags/TagDelete';
+
+export default {
+	props: {
+		id: {
+			type: String|Number,
+			required: true,
+		},
+	},
+	data: () => {
```
bognix: `data() {}`

app/Models/Tag.php
rogatty: FYI we discussed it already in https://github.com/bethinkpl/wnl-platform/pull/1192#discussion_r242436492 and @kvas-damian decided to override the `delete` method as it's easier.

Now I'm thinking this might be a bad idea, as we open a DB transaction for as long as everything in the parent's `delete` runs, e.g. multiple events are fired.

resources/assets/js/admin/components/tags/TaggablesMover.vue
```diff
@@ -0,0 +1,57 @@
...
+<script>
+import {mapActions} from 'vuex';
+import {Form as WnlForm, Text as WnlFormText, Submit as WnlSubmit} from 'js/components/global/form';
+
+export default {
+	name: 'TaggablesMover',
+	props: {
+		beforeSubmit: {
+			type: Function,
+			default: null,
```
bognix: maybe noop function is better as default? To avoid calling null by accident?

app/Models/Tag.php
bognix: I see the benefit of having a transaction and observables are called one after another so the probability of broken state is much higher than in case of transaction. I leave this decision to you :) I see you thought it through.

kvas-damian: I think this is a topic for a bigger discussion. 
DB tansaction ensures consistency, we could try to hack the method to make sure that Observables are called inside the transaction, but then we need to change approach to observables. Currently most of them are queuing job to the queue - this approach can't be connected to the transaction, so we loose the data integrity. This is why I decided to not use observers.
Another approach  could be that if the job in the queue fails, we can get back to it and manually check what happened and resolve the problem. I believe such problems could very rare, but still... requires manual work.

rogatty: @kvas-damian are we sure that nothing in `parent::delete()` calls external services? A few days ago I got a transaction timeout because of connectivity issues with Elasticsearch. If we're calling it synchronously, then putting a transaction here might be a bad idea.

resources/assets/js/admin/components/tags/TaggablesMover.vue
bognix: I prefer a bit stricter approach to typing. If something is expected to be a function, let it be a function, not null. Adding noop here is not complicated and you don't need to check it in the Form.vue

resources/assets/js/admin/components/tags/TagDelete.vue
kvas-damian: I'd go opposite way. I believe it isn't modal responsibility to decide if it has to be shown or not, that's why I would migrate to use `v-if` only.
This solution also has a positive side-effect on performance: you don't need to instantiate a component to decide that it should be hidden.

What do you think?

bognix: makes sense :+1:

app/Models/Tag.php
kvas-damian: Scout is using a queue for synching, so elasticsearch isn't a problem here. But the queue is also an external service.
What do we do about it?

kvas-damian: No!

PR #1193, opened 2018-12-18
PLAT-826 | Tags list in admin

kvas-damian:
Highlights:
* Frontend: introduce `PaginatedList` component for rendering paginated lists with search (using `.filter` endpoint)
* Frontend: introduce `SortableTable` component for rendering paginated sortable lists (using `.filter` endpoint, based on `PaginatedList`)
* Backend: call `withCount()` on a model when counts are present in `include` query param. Example: `?include=taggables_count`. Then count results are available in `ApiTransformer`.


resources/assets/js/admin/components/lists/components/SortableTableColumnHeader.vue
```diff
@@ -0,0 +1,32 @@
...
+	<th @click="$emit('click')" class="columnName">
+		{{label}}
+		<span class="icon is-small" v-if="isActive">
+			<i :class="['fa', sortDirection === 'asc' ? 'fa-chevron-up': 'fa-chevron-down']"></i>
+		</span>
+	</th>
+</template>
+
+<style lang="sass">
+	.columnName
```
bognix: classes with dash

app/Http/Controllers/Api/Transformers/TaggablesCountTransformer.php
```diff
@@ -0,0 +1,33 @@
...
+
+	public function __construct($parent = null)
+	{
+		$this->parent = $parent;
+	}
+
+	public function transform(Tag $tag)
+	{
+		$data = [
+			'taggables_count'   => $tag->taggables_count,
```
bognix: dziwne spacje

resources/assets/js/admin/components/slides/annotations/Annotations.vue
```diff
@@ -77,34 +66,26 @@
...
 <script>
-import axios from 'axios';
 import {mapActions} from 'vuex';
-
-import { getApiUrl } from 'js/utils/env';
-import AnnotationsList from './AnnotationsList';
-import AnnotationsEditor from './AnnotationsEditor';
-import Pagination from 'js/components/global/Pagination';
-import WnlSearchInput from 'js/components/global/SearchInput';
+import WnlAnnotationsList from './AnnotationsList';
```
bognix: this.annotations is no longer used. Figure out how to refresh the annotations list without refetching the items.

resources/assets/js/admin/components/tags/TagsList.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+	<div class="orders-list">
+		<wnl-sortable-table
+			:resourceUrl="getApiUrl('tags/.filter')"
```
bognix: getApiUrl should be inside PaginatedList, here we should provide resource-route.

app/Http/Controllers/Api/ApiController.php
```diff
@@ -238,6 +241,22 @@ protected function eagerLoadIncludes(string $model)
...
+		if (empty($this->include)) {
+			return $model;
+		}
+
+		$withCountIncludes = array_filter(explode(',', $this->include), function($include) {
+			return Str::endsWith($include, self::WITH_COUNT_INCLUDE_SUFFIX);
+		});
+
+		$model->withCount(array_map(function($include) {
+			return str_replace(self::WITH_COUNT_INCLUDE_SUFFIX, '', $include);
```
rogatty: It'll produce incorrect result for `x_count_y_count` but I don't think we need to worry about that.

app/Http/Controllers/Api/Transformers/TagTransformer.php
```diff
@@ -11,6 +11,9 @@
 class TagTransformer extends ApiTransformer
 {
 	protected $parent;
+	protected $availableIncludes = [
+		'taggables_count'
```
rogatty: Shouldn't we have either the `_count` handled in `loadCountIncludes` or this explicit include here? Do we need both?

app/Http/Controllers/Api/Transformers/TaggablesCountTransformer.php
```diff
@@ -0,0 +1,33 @@
...
+class TaggablesCountTransformer extends ApiTransformer
+{
+	protected $parent;
+
+	public function __construct($parent = null)
+	{
+		$this->parent = $parent;
+	}
+
+	public function transform(Tag $tag)
```
rogatty: If this transforms a `Tag` then the `TaggablesCount` name might be misleading. How about `TagTaggablesCount`?

app/Http/Controllers/Api/Transformers/TagTransformer.php
kvas-damian: I think we need both. The one in controller just makes a query which includes counts in the result. This is the resource transformer which need to know if the count should be included in the response or not.

app/Http/Controllers/Api/Transformers/TaggablesCountTransformer.php
kvas-damian: What do you think about making this transformer tag independent? It could be used on other resources that have a relation to taggables.

resources/assets/js/admin/components/Sidenav.vue
```diff
@@ -14,15 +14,14 @@
 		</p>
 		<ul class="menu-list">
 			<li><router-link :to="{ name: 'slides' }">Edytuj Slajd</router-link></li>
-		</ul>
-		<ul class="menu-list">
 			<li><router-link :to="{ name: 'add-slide' }">Dodaj Slajd</router-link></li>
-		</ul>
-		<ul class="menu-list">
 			<li><router-link :to="{ name: 'charts' }">Diagramy</router-link></li>
+			<li><router-link to="/admin/app/annotations">Przypisy</router-link></li>
```
rogatty: Why not `:to="{ name: 'annotations' }"`?

resources/assets/js/admin/components/Sidenav.vue
```diff
@@ -14,15 +14,14 @@
...
 		</ul>
+
+		<p class="menu-label">Tagi</p>
 		<ul class="menu-list">
-			<li><router-link to="/admin/app/annotations">Przypisy</router-link></li>
+			<li><router-link :to="{ name: 'tags' }">Tagi</router-link></li>
```
rogatty: Looking at the items above, I think it should be `Edycja tagów`

resources/assets/js/admin/components/flashcards/list/FlashcardsList.vue
```diff
@@ -1,98 +1,40 @@
...
-		<div v-if="!isLoading">
-			<wnl-flashcard-list-item v-for="flashcard in flashcards"
-															 :key="flashcard.id"
-															 :content="flashcard.content"
-															 :id="flashcard.id"
-			/>
-		</div>
-		<wnl-text-loader v-else></wnl-text-loader>
+		<wnl-paginated-list
+			:resource-name="'flashcards/.filter'"
```
rogatty: This is the recommended style based on https://vuejs.github.io/eslint-plugin-vue/rules/attribute-hyphenation.html but we use camelCase everywhere I looked. Probably not something to worry about, I'll convert it at some point 😛

kvas-damian: I trust you :)

resources/assets/js/admin/components/lessons/Lessons.vue
```diff
@@ -36,16 +35,9 @@ export default {
...
-		...mapGetters('lessons', ['isReady']),
 		lessonId() {
 			return this.$route.params.lessonId;
 		},
 	},
-	methods: {
-		...mapActions('lessons', ['setup']),
-	},
-	mounted() {
-		this.setup();
```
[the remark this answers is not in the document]
kvas-damian: We use data from `PaginatedList` component, so we don't need to setup store. The store is still used in `FlashcardsSetEditor` so we can't delete it.

resources/assets/js/admin/components/lists/ApiSortableTable.vue
```diff
@@ -0,0 +1,77 @@
+<template>
```
rogatty: I'd call this component `PaginatedSortableTable`

resources/assets/js/admin/components/lists/PaginatedList.vue
```diff
@@ -0,0 +1,160 @@
...
+			await this.fetch();
+		},
+		async fetch() {
+			try {
+				this.isLoading = true;
+				if (this.requestCancelTokenSource) {
+					// cancel previous request
+					this.requestCancelTokenSource.cancel();
+				}
+				this.requestCancelTokenSource = axios.CancelToken.source();
```
rogatty: This is never set to `null`, isn't it a problem?

kvas-damian: Ithink it's not a problem.

resources/assets/js/admin/components/lists/SortableTable.vue
```diff
@@ -0,0 +1,90 @@
...
+				sortDirection,
+				activeSortColumnName: name,
+			});
+
+			this.$router.push({ query: { ...this.$route.query, sortDirection, sort: name }});
+		},
+	},
+	mounted() {
+		if (this.routerSort && (this.routerSortDirection !== this.sortDirection || this.routerSort !== this.activeSortColumnName)) {
+			this.$emit('changeOrder', {
```
rogatty: If user loads the page with sort query params, we'll load the model twice and there will be a flash. We should fix it before using this component in regular user facing places.

rogatty:
Will be merged to master as a part of https://github.com/bethinkpl/wnl-platform/pull/1192

PR #1194, opened 2018-12-18
PLAT-838 | Fix exam

kvas-damian:


resources/assets/js/components/course/screens/MockExam.vue
```diff
@@ -21,6 +21,7 @@
 
 <script>
 import _ from 'lodash';
```
bognix: js/consts/quetionsSolving please

PR #1197, opened 2018-12-20
PLAT-851 | Show related taggables list in tag editor

rogatty:
https://bethink.atlassian.net/browse/PLAT-851

### DIFF
--- a/app/Http/Controllers/Api/Filters/TaggableFilter.php
+++ b/app/Http/Controllers/Api/Filters/TaggableFilter.php
@@ -1,0 +1,16 @@
+<?php namespace App\Http\Controllers\Api\Filters;
+
+class TaggableFilter extends ApiFilter
+{
+	protected $expected = ['tag_id'];
+
+	public function handle($builder)
+	{
+		return $this->collection($builder);
+	}
+
+	protected function collection($builder)
+	{
+		return $builder->where('tag_id', $this->params['tag_id']);
+	}
+}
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/Api/Filters/TaggableFilter.php
```diff
@@ -0,0 +1,20 @@
+<?php namespace App\Http\Controllers\Api\Filters;
+
+class TaggableFilter extends ApiFilter
+{
+	protected $expected = ['tag_id'];
+
+	public function handle($builder)
```
rogatty:
```suggestion
 	public function handle(Builder $builder)
```

app/Http/Controllers/Api/Filters/TaggableFilter.php
```diff
@@ -0,0 +1,20 @@
...
+	{
+		return $this->collection($builder);
+	}
+
+	protected function collection($builder)
```
rogatty:
```suggestion
	protected function collection(Builder $builder)
```

resources/assets/js/admin/components/tags/TagEditor.vue
```diff
@@ -117,6 +213,19 @@ export default {
 				is_delete_allowed: get(formData, `included.meta.${formData.id}.is_delete_allowed`, false),
 				is_rename_allowed: get(formData, `included.meta.${formData.id}.is_rename_allowed`, true),
 			};
+		},
+		getTaggableLink({taggable_type, taggable_id}) {
+			switch (taggable_type) {
+			case 'App\\Models\\QuizQuestion':
```
bognix: reuse the map we already have

PR #1198, opened 2018-12-20
Fix paginated response caching

gitkuba:


bognix:
In my opinion instead of cloning model we shouldn't use cached pagination in case of static paginated lists. CachedPagination introduces a lot of complexity and was developed to solve one particular problem - the solved/unsolved filter in questions bank. What is more CachedPagination has drawbacks and doesn't always work perfectly, it would be great if we could revisit the problem once again and think of better solution. However, until this time comes we shouldn't use this pagination for any other lists to minimize the scope of potential refactor in the future

PR #1199, opened 2018-12-20
[PLAT-767] admin can browse, search, edit and create new question

preziu:
- [ ] add elastic search indexing for QuizQuestion on production


resources/assets/js/admin/components/quizes/list/QuizQuestionsList.vue
```diff
@@ -0,0 +1,40 @@
...
+</template>
+
+<script>
+import WnlQuizQuestionListItem from 'js/admin/components/quizes/list/QuizQuestionListItem';
+import WnlPaginatedList from 'js/admin/components/lists/PaginatedList';
+
+export default {
+	name: 'QuizQuestionsList',
+	components: {
+		'quiz-question-list-item': WnlQuizQuestionListItem,
```
kvas-damian: Can we try to use `wnl` prefix for all our components?

resources/assets/js/admin/components/quizes/list/QuizQuestionListItem.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+	<div class="margin bottom item">
+		<router-link :to="to" class="link">
+			<span class="id">{{id}}. </span>
+			<span class="text" v-html="text" />
```
kvas-damian: What do you think about displaying first XX chars of the question? So questions are really long and we don't need entire question content on the list.
Also we could consider stripping HTML tags - currently it looks odd when you have some questions styled on the list.

resources/assets/js/admin/components/quizes/list/QuizQuestionsList.vue
preziu: Good point! I'll implement that convention for future (and this one) custom components, cheers!

PR #1201, opened 2019-01-02
PLAT-863 | Enable tag deletion

rogatty:
https://bethink.atlassian.net/browse/PLAT-863
Follow up to https://github.com/bethinkpl/wnl-platform/pull/1192

### DIFF
--- a/app/Http/Controllers/Api/PrivateApi/TaggablesApiController.php
+++ b/app/Http/Controllers/Api/PrivateApi/TaggablesApiController.php
@@ -4,7 +4,9 @@ use App\Http\Controllers\Api\ApiController;
 use App\Http\Requests\Course\MoveTaggable;
 use App\Models\Tag;
 use App\Models\Taggable;
+use Illuminate\Database\Query\Builder;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\DB;
 
 class TaggablesApiController extends ApiController {
 	public function __construct(Request $request) {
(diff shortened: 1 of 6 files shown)


app/Models/Tag.php
```diff
@@ -32,37 +33,47 @@ public function taxonomies()
...
-	public function isRenameAllowed() {
-		// TODO
-		return false;
-		// SlidesFromCategory command uses hardcoded tag names
-		 return !$this->isCategoryTag();
+	public function isRenameAllowed()
+	{
+		// `Prezentacja` tag doesn't have a protected taggable but needs to be protected anyway
+		// It is used to display QnA below presentation but not intro, quiz questions etc.
+		return $this->name !== 'Prezentacja' &&
```
kvas-damian: Please use String constant

app/Http/Controllers/Api/PrivateApi/TaggablesApiController.php
```diff
@@ -23,22 +25,38 @@ public function batchMove(MoveTaggable $request, $sourceTagId) {
...
-			->update(['tag_id' => $targetTag->id]);
-
-		return $this->respondOk([
-			'taggables_moved' => $taggablesMoved
-		]);
+		$transactionResult = DB::transaction(function () use ($targetTag, $sourceTag) {
+			$taggablesDeleted = Taggable::where('tag_id', $sourceTag->id)
+				->whereExists(function ($query) use ($targetTag) {
+					/** @var $query Builder */
+					$query->select(DB::raw(1))
```
kvas-damian: this solution performs a subquery for every row with `sourceTag`. Instead, you can select all ids you want to delete in one query and put it into where:
`delete from taggables where id in (select id from (select t1.id from taggables t1 join taggables t2 on t1.taggable_id = t2.taggable_id and t1.taggable_type = t2.taggable_type where t1.tag_id = 28 and t2.tag_id = 39) as tmp);`

PR #1202, opened 2019-01-02
PLAT-792 | Upgrade dependencies

rogatty:
- Upgrade laravel-mix explicitly and babel implicitly.
- Remove obsolete webpack config and build utils.
- Remove `lang="html"` from Vue templates, see https://github.com/vuejs/vue-loader/issues/1443.
- Use static `import` instead of `require()` wherever possible (https://laravel-mix.com/docs/4.0/upgrade#importing-es-modules). In some places we left `require()` because these places would need a dynamic `import()` which turned out problematic. After enabling dynamic `import()` (https://github.com/bethinkpl/wnl-platform/pull/1202/commits/054bd0ea0b06433558231a2c6e08b51a344d3ed3) the content of a compiled `app.css` was `/* (ignored) */` instead of the actual CSS. If we ever want to get rid of `require()`, we need to fix this issue first.
- Import slideshow styles using JS instead of regular CSS files. In https://github.com/JeffreyWay/laravel-mix/pull/1743 laravel-mix was changed to use `style-loader` when HMR is active. Because of that, `slideshow.css` was loaded in the main app, instead of only in the slideshow iframe. We could either fix the styles to be non-conflicting or load them using `import` in JS. The latter was much cheaper and consistent with how we load the Vue styles.
- Use laravel-mix fork, PR to the upstream pending: https://github.com/JeffreyWay/laravel-mix/pull/1919
- Hide `notSupportedBrowserModal` before JS is loaded. It was originally hidden by Bulma styles which are now loaded with JS.
- Use Chrome headless instead of PhantomJS in JS tests.
- Manually copy `resources/assets/images/emoji.png` from `public/` so it's accessible from `resources/vendor/emoji/emoji.scss`.

TODO:
- [x] Fix 404 for fonts http://platform.local/fonts/H2DMvhDLycM56KNuAtbJYA.woff2?b68c39cb688a8efff6f1c19f55313dd2 and http://platform.local/fonts/UyBMtLsHKBKXelqf4x7VRQ.woff2?33f84b19c6dad2fa605bfea63564b5cb
- [x] Migrate from require to imports https://webpack.js.org/migrate/4/#import-and-commonjs (example: https://github.com/bethinkpl/wnl-platform/blob/90f1545c01e004ea2acec7b48f0b5cb57f4e3b23/resources/assets/js/router.js#L17)
- [x] Fix admin app
- [x] Fix datepicker: https://github.com/bethinkpl/wnl-platform/blob/90f1545c01e004ea2acec7b48f0b5cb57f4e3b23/resources/assets/js/components/global/form/Datepicker.vue#L90
- [x] Fix `.slideshow-annotations` div positioning (similar problem with bookmark)
- [x] Fix js unit tests
- [x] Fix styles for search in admin panel lists


resources/assets/js/plugins/axios.js
```diff
@@ -8,7 +9,7 @@ export default (Vue, {store, router}) => {
 	 * included with Laravel will automatically verify the header's value.
 	 */
 
-	window.axios = require('axios');
+	window.axios = axios;
```
bognix: _note_: check if comments are not added/displayed twice when adding one below slide.

webpack.mix.js
```diff
@@ -7,31 +7,111 @@ const {mix} = require('laravel-mix');
...
 
 if (mix.config.inProduction) {
-	mix.version()
+	mix.version();
 }
 
 if (process.env.SYNC === 'on') {
 	mix.browserSync('platforma.wnl');
 }
+
```
bognix: the if above is not needed as well I think. We no longer have 'platfroma.wnl' URL on dev.

resources/assets/js/plugins/axios.js
rogatty: They're not.

kvas-damian:
`resources/assets/images/emoji.png` Do we have this file in places now?

kvas-damian:
`app.scss` isn't loaded when running `yarn run dev`.

resources/assets/sass/app.scss
```diff
@@ -14,6 +14,7 @@
 @import "./views/login";
 
 // Fonts
+$fa-font-path: "../../../node_modules/font-awesome/fonts";
```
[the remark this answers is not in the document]
bognix: This is required by bulma.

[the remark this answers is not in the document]
rogatty: Yup.

PR #1203, opened 2019-01-03
Fix paginated response cache

gitkuba:


kvas-damian:
Fix unittests

app/Http/Controllers/Api/Concerns/ProvidesApiFiltering.php
```diff
@@ -42,17 +43,18 @@ public function filter(Request $request)
...
+		} elseif ($cachedPagination) {
+			// Cached pagination is meant to be used within quiz questions bank only, as it was
+			// developed to address a specific issue while working with solved/unsolved filter.
+			// To be refactored, see https://bethink.atlassian.net/browse/PLAT-868
+			$cacheTags = $this->getFiltersCacheTags($this->resourceName, $userFiltersPersistanceToken);
+			$hashedFilters = $this->hashedFilters($filters);
+
+			$response = $this->cachedPaginatedResponse($cacheTags, $hashedFilters, $model, $this->limit, $this->page);
 		} else {
-			if (!$request->has('active') || empty($filters)) {
```
kvas-damian: Previously we didn't cache for these conditions. Are you sure we want to change that for `_fetchQuestions` method?

gitkuba: If I correctly understood previous implementation, checking for presence of `active` or `filters` was a way to determine the request is from questions bank, because it is (was) the only place, in which we use active filters.

kvas-damian: Previously we're sending:
```
active: []
filters: []
```
Which means no cache. Now we send:
```
active: []
cachedPagination: true
filters: []
```
which means use case - so we change the behavior. The question is if the change was intended?

app/Http/Controllers/Api/Concerns/ProvidesApiFiltering.php
```diff
@@ -42,17 +44,18 @@ public function filter(Request $request)
 
 		$model = $this->loadCountIncludes($model);
 
-		if (!empty($randomize)) {
+		if ($randomize) {
 			$response = $this->randomizedResponse($model, $this->limit);
+		} elseif ($cachedPagination && ($active || $filters)) {
```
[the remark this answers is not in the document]
gitkuba: Trust me, I'm an engineer ;)

PR #1208, opened 2019-01-04
PLAT-865 | detach Qna from tags

bognix:


app/Console/Commands/MigrateQna.php
```diff
@@ -0,0 +1,111 @@
...
+	public function handle()
+	{
+		$qnaQuestions = QnaQuestion::get();
+
+		$bar = $this->output->createProgressBar($qnaQuestions->count());
+		$bar->start();
+
+		foreach ($qnaQuestions as $qnaQuestion) {
+			$bar->advance();
+			$qnaQuestion = QnaQuestion::find($qnaQuestion->id);
```
bognix: this one I don't like - but I have to refresh the model because the discussion could already be added.

app/Providers/DiscussableServiceProvider.php
```diff
@@ -0,0 +1,18 @@
...
+
+namespace App\Providers;
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\ServiceProvider;
+
+class DiscussableServiceProvider extends ServiceProvider {
+	public function register() {
+		Blueprint::macro('discussable', function() {
+			$this->tinyInteger('discussable')->default(0);
```
kvas-damian: maybe `is_discussable`

app/Console/Commands/MigrateQna.php
```diff
@@ -0,0 +1,107 @@
...
+			$matchingQnaQuestions = $qnaQuestionsQuery->get();
+
+			$screensQuery = Screen::select();
+			foreach ($tags as $tagId) {
+				$screensQuery->whereHas('tags', function ($query) use ($tagId) {
+					$query->where('tags.id', $tagId);
+				});
+			}
+			$name = '';
+			$matchingDiscussable = $screensQuery->first();
```
rogatty: Is it possible that `$screensQuery` returns more than one row? If so, should we at least log it?

rogatty: Is it possible that `$screensQuery` returns more than one row? If so, should we at least log it?

app/Console/Commands/MigrateQna.php
```diff
@@ -0,0 +1,107 @@
...
+				foreach ($tags as $tagId) {
+					$pagesQuery->whereHas('tags', function ($query) use ($tagId) {
+						$query->where('tags.id', $tagId);
+					});
+				}
+
+				$matchingDiscussable = $pagesQuery->first();
+
+				if (empty($matchingDiscussable)) {
+					continue;
```
rogatty: How about logging and then cleaning up the QnAs which aren't displayed anywhere?

app/Console/Commands/MigrateQna.php
```diff
@@ -0,0 +1,107 @@
...
+				} else {
+					$name = "{$matchingDiscussable->slug}";
+				}
+			} else {
+				// qna questions have matching screen
+				$name = "{$matchingDiscussable->lesson->name} - {$matchingDiscussable->name}";
```
rogatty: This will be out of date after lesson / screen name change. Do we plan to keep them in sync?

app/Console/Commands/MigrateQna.php
```diff
@@ -0,0 +1,107 @@
...
+			$qnaQuestion = QnaQuestion::find($qnaQuestion->id);
+
+			if (!empty($qnaQuestion->discussion_id)) {
+				// if question is already attached to discussion, skip it
+				continue;
+			}
+
+			$tags = $qnaQuestion->tags->pluck('id');
+			if ($tags->count() === 0) {
+				$this->output->warning("Qna Question without tags: {$qnaQuestion->id}");
```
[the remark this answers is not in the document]
bognix: yes we do... I removed them after all but as part as the second subtask in this story (the frontend work)

app/Console/Commands/MigrateQna.php
bognix: no, I double checked with production dump. Creating logic for handling case for more than one matching discussable was redundant given it's a "one time" script

app/Console/Commands/MigrateQna.php
bognix: it's going to happen as part as the second subtask. I already put the query that we should run after making the data migration and before releasing the second subtask (it's in PLAT-873)

app/Console/Commands/MigrateQna.php
bognix: no, it's just for the purpose of the migration. In the future when screen is created admin can choose whether this screen is discussable or not. If it's discussable then the name can be provided.

Right now I wanted to add some descriptive names that if for some reason we loose the connection between the screen and discussion we can deduct where the discussion should be attached based on this name.

resources/assets/js/components/qna/NewQuestionForm.vue
```diff
@@ -61,15 +61,25 @@ export default {
...
 	},
-	props: ['tags'],
+	props: {
+		tags: {
+			type: Array,
+			default: []
+		},
+		discussionId: {
+			type: Number,
+			default: 0
```
kvas-damian: Is it really ok if we pass 0 to backend? I think it should be required.

PR #1212, opened 2019-01-09
Give queue worker containers explicit names

gitkuba:


kvas-damian:
Can you give us some background behind these changes?

gitkuba:
@kvas-damian Names of containers with job queues are confusing and you have to look into `docker-compose.yml' to tell which is the one you're interested in. So i've changed container names, so i could more easily check out the logs or restart the desired queue.

PR #1215, opened 2019-01-10
PLAT-867 | TaxonomyTerms structure

kvas-damian:
* using https://github.com/lazychaser/laravel-nestedset for tree storage and manipulations


app/Http/Controllers/Api/PrivateApi/TaxonomyTermsApiController.php
```diff
@@ -0,0 +1,16 @@
...
+use Illuminate\Http\Request;
+
+class TaxonomyTermsApiController extends ApiController {
+	public function __construct(Request $request) {
+		parent::__construct($request);
+		$this->resourceName = config('papi.resources.taxonomy-terms');
+	}
+
+	public function getByTaxonomy($taxonomyId) {
+		return $this->transformAndRespond(TaxonomyTerm::where('taxonomy_id', $taxonomyId)->get()->toFlatTree());
```
kvas-damian: Currently, I return a flatTree (collection with `parentId` attribute) to the frontend and build the tree there. This allows us to use standard Fractal transformers and includes.
The alternative solution is to return a nested structure, which makes frontend easier, although I didn't manage to make Fractal work with nested resources correctly (there is a problem with collecting includes from all levels). Another solution would be to use a custom structure that bypasses Fractal Transformers, although at this moment I was afraid of making this endpoint exceptional.

What are your thoughts? Do you see any alternatives? Maybe you have some suggestions?

resources/assets/js/admin/components/taxonomies/TaxonomyTermItem.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+	<li>
+		{{ term.id }}. {{included.tags[term.tags[0]].name}}
+		<ul v-if="childTerms.length">
+			<wnl-taxonomy-term-item
+				v-for="childTerm in childTerms"
+				:key="childTerm.id"
+				:term="childTerm"
+				:included="included"
+				:terms="terms"
```
rogatty: Consider using `left` and `right` to minimize the list size that we pass here.

kvas-damian: We agreed offline that we'd like to apply performance optimizations after we migrate to vuex

rogatty: Let's re-evaluate when we have the state in Vuex.

PR #1225, opened 2019-01-14
log to Sentry when X-Socket-ID header is not send to backend

bognix:


app/Events/Event.php
```diff
@@ -25,6 +25,11 @@ public function __construct()
 	{
 		$this->referer = Request::header('X-BETHINK-LOCATION');
 		$this->id = Uuid::uuid4()->toString();
+
+		if (!Request::header('X-Socket-ID')) {
+			$sentryClient = new \Raven_Client(env('SENTRY_DSN'));
+			$sentryClient->captureException(new \Exception('X-Socket-ID header missing'));
```
gitkuba: Please make it on prod only or use `Log::error()` instead of directly sending to sentry.

bognix: Make sense, we don't send any additional context so Log::error should be sufficient.

PR #1227, opened 2019-01-15
[PLAT-891] Test user

gitkuba:


app/Console/Commands/UserClone.php
```diff
@@ -0,0 +1,152 @@
...
+			die;
+		}
+
+		$this->copyRedisData($redis, $targetUser, $sourceUser, $courseId);
+
+		foreach (self::$TABLES as $table) {
+			$this->copyTable($table, $sourceUser, $targetUser);
+		}
+
+		$targetUser->subscription()->update(['access_end' => Carbon::now()->addYears(15)]);
```
bognix: can we determine the access end date based on the presence of test role?

I think we agreed that any modifications to access should be handled by role

bognix: @kvas-damian I wonder what is your opinion :)

app/Console/Commands/UserClone.php
```diff
@@ -0,0 +1,152 @@
...
+		'exams_results',
+		'reactables',
+		'role_user',
+		'user_quiz_results',
+		'user_settings',
+		'users_plan_progress',
+		'users_plans',
+		'user_course_progress',
+		'user_lesson',
+		'user_subscription',
```
bognix: shouldn't we copy orders, instalments (if present) as well?

docker-compose.yaml
```diff
@@ -17,13 +17,15 @@ services:
 
   redis:
     container_name: redis
-    image: redis:4.0.9-alpine
+    image: redis:5.0.3-alpine
```
bognix: what redis version we use on production?

app/Console/Commands/UserClone.php
```diff
@@ -0,0 +1,152 @@
...
+			$insert = $results->map(function ($row) use ($targetUser) {
+				$row->user_id = $targetUser->id;
+				unset($row->id);
+				return (array)$row;
+			})->toArray();
+
+			foreach ($insert as $item) {
+				try {
+					\DB::table($table)->insert($item);
+				} catch (\Illuminate\Database\QueryException $e) {
```
bognix: This thing can only happen if we copy data to existing user, not creating new one?

app/Console/Commands/UserClone.php
```diff
@@ -0,0 +1,152 @@
...
+	private function copyTable($table, $sourceUser, $targetUser)
+	{
+		\DB::beginTransaction();
+
+		try {
+			$results = \DB::table($table)->where('user_id', $sourceUser->id)->get();
+			$insert = $results->map(function ($row) use ($targetUser) {
+				$row->user_id = $targetUser->id;
+				unset($row->id);
+				return (array)$row;
```
bognix: according to the docs, `toArray` converts nested object to array as well (https://laravel.com/docs/5.7/collections#method-toarray) - there is no need to cast it to array here.

docker-compose.yaml
gitkuba: 5.0.3

app/Console/Commands/UserClone.php
gitkuba: Yup, but command actually allows You to copy data to an existing user. I thought it'd be a nice feature :)

app/Console/Commands/UserClone.php
gitkuba: Unluckily it didn't work for me 🤷‍♂️

app/Console/Commands/UserClone.php
bognix: Ok, then lesson planner won't fully work on this person. Wasn't the goal to have a user with fully functional platform?

PR #1229, opened 2019-01-16
PLAT-888 | Sort terms

bognix:


resources/assets/js/admin/components/taxonomies/TaxonomyTermItem.vue
```diff
@@ -88,11 +99,15 @@ export default {
...
+				.filter(term => term.parent_id === this.term.id)
+				.sort((termA, termB) => termA.orderNumber - termB.orderNumber);
 		},
-		...mapGetters('taxonomyTerms', ['filteredTerms']),
 	},
-	// Name is required to allow recursive rendering
-	name: 'wnl-taxonomy-term-item',
+	methods: {
+		move(...args) {
+			this.$emit('moveTerm', ...args);
```
kvas-damian: Maybe we could call this method on store directly instead of bubbling?

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -42,6 +42,28 @@ const mutations = {
 	[types.SET_TAXONOMY_TERMS_FILTER] (state, payload) {
 		set(state, 'filter', payload);
 	},
+	[types.MOVE_TERM] (state, {term, replaceWith}) {
+		const replaceWithOrderNumber = replaceWith.orderNumber;
+		const termOrderNumber = term.orderNumber;
+
+		set(state, 'terms', state.terms.map(stateTerm => {
```
kvas-damian: vuex prefers mutations

app/Http/Controllers/Api/PrivateApi/TaxonomyTermsApiController.php
```diff
@@ -46,4 +51,17 @@ public function put(UpdateTaxonomyTerm $request) {
 
 		return $this->transformAndRespond($taxonomyTerm);
 	}
+
+	public function move(MoveTaxonomyTerm $request) {
+		$target = TaxonomyTerm::find($request->get('term_id'));
+		$direction = $request->get('direction');
+
```
bognix: if direction is 0 we shouldn't perform any operations.

resources/assets/js/admin/components/taxonomies/TaxonomyTermItem.vue
```diff
@@ -118,8 +136,20 @@ export default {
 				this.expandTerm(this.term.id);
 			}
 		},
+		onChildTermMove({term, direction}) {
+			const oldIndex = this.rootTerms.indexOf(term);
```
bognix: `this.childTerms.indexOf` - how come it works.

app/Http/Controllers/Api/PrivateApi/TaxonomyTermsApiController.php
```diff
@@ -46,4 +51,17 @@ public function put(UpdateTaxonomyTerm $request) {
...
+		if ($direction > 0) {
+			$target->down($direction);
```
kvas-damian: Let's check what will happen when move the item too much. `$direction=-1000`

resources/assets/js/admin/components/taxonomies/TaxonomyTermItem.vue
```diff
@@ -71,10 +81,16 @@
...
-import {mapActions, mapGetters, mapState} from 'vuex';
+import {mapActions, mapState} from 'vuex';
 import {TAXONOMY_EDITOR_MODES} from 'js/consts/taxonomyTerms';
+import draggable from 'vuedraggable';
 
 export default {
+	// Name is required to allow recursive rendering
+	name: 'wnl-taxonomy-term-item',
+	components: {
+		draggable
```
kvas-damian: https://vuejs.org/v2/guide/components-registration.html
> we strongly recommend following the W3C rules for custom tag names (all-lowercase, must contain a hyphen)

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -104,6 +138,31 @@ const actions = {
 		commit(types.SET_TAXONOMY_TERMS_SAVING, false);
 	},
 
+	async dragTerm({commit, dispatch}, {terms, oldIndex, newIndex}) {
```
kvas-damian: I'm not sure if `dragTerm` is an intuitive name for this action on the store. Maybe we should call it `moveTo`?

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -104,6 +138,31 @@ const actions = {
 		commit(types.SET_TAXONOMY_TERMS_SAVING, false);
```
kvas-damian: Make sure we have the right order after update and create

kvas-damian:
As we discussed offline, please think how to avoid duplicate code for move/drag operations.

resources/assets/js/admin/components/taxonomies/TaxonomyTermItem.vue
```diff
@@ -118,8 +125,12 @@ export default {
...
 			}
 		},
+		onTermMove(term, direction) {
+			this.$emit('moveTerm', {term, direction});
+		},
 	},
-	// Name is required to allow recursive rendering
-	name: 'wnl-taxonomy-term-item',
+	beforeCreate: function () {
+		this.$options.components.WnlTaxonomyTermsList = require('./TaxonomyTermsList.vue').default;
```
bognix: @kvas-damian this is crazy. I'm not sure if it's really better option with this components layout.
https://vuejs.org/v2/guide/components-edge-cases.html#Circular-References-Between-Components

kvas-damian: agree, it looks like a hack. I'm even more surprised it's a suggested way of solving such problems :(

kvas-damian: Actually it's hard to decide which solution is better, but I'm more with the separated structure. This is the only line that requires explanation. In the previous solution it was hard to follow the action flow.

Summarizing, I'd add a hack description here with the link and leave it as it is.

app/Http/Controllers/Api/PrivateApi/TaxonomyTermsApiController.php
bognix: It won't succeed and the library returns false. I added the support for this case.

resources/assets/js/admin/mixins/scroll-to-taxonomy-term.js
```diff
@@ -3,10 +3,16 @@ import {scrollToElement} from 'js/utils/animations';
...
 			await Vue.nextTick();
 			// wait for collapse animation to finish
-			setTimeout(() => scrollToElement(document.getElementById(`term-${term.id}`), 150, 500, document.querySelector('.admin-right')), 300);
+			setTimeout(
+				() => scrollToElement(
+					document.getElementById(`term-${term.id}`),
+					150,
+					500,
+					document.querySelector('.admin-right')
+				), 300);
```
bognix: indent

resources/assets/js/admin/components/taxonomies/TaxonomyTermEditorForm.vue
```diff
@@ -79,20 +79,23 @@ export default {
...
+
 			this.onSave({
 				id: this.id,
-				parent_id: this.parent ? this.parent.id : null,
+				parent_id: parentId,
 				tag_id: this.tag.id,
 				description: this.description,
-				taxonomy_id: this.taxonomyId
+				taxonomy_id: this.taxonomyId,
+				orderNumber: this.getSiblingsByParentId({parentId, id: this.id}).length
```
bognix: only update order number in case changing the parent

resources/assets/js/admin/components/taxonomies/TaxonomyTermItem.vue
```diff
@@ -45,6 +48,7 @@
 
 	.taxonomy-term-item
 		&__content
+			cursor: move
```
kvas-damian: maybe we could add an icon (three dots) indicating that item can be moved?

resources/assets/js/admin/components/taxonomies/TaxonomyTermsList.vue
```diff
@@ -0,0 +1,56 @@
...
+</template>
+
+<script>
+import {mapActions} from 'vuex';
+
+import VueDraggable from 'vuedraggable';
+import WnlTaxonomyTermItem from 'js/admin/components/taxonomies/TaxonomyTermItem';
+
+export default {
+	// Name is required to allow recursive rendering
```
kvas-damian: I think it's not required after your changes

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -110,14 +148,39 @@ const actions = {
...
+		if (direction === 0) return;
+
+		const term = terms[oldIndex];
+		terms.splice(oldIndex, 1);
+		terms.splice(newIndex, 0, term);
+
+		commit(types.REORDER_TERMS, terms);
+
+		try {
+			await axios.put(getApiUrl('taxonomy_terms/move'), {
```
kvas-damian: to consider, we can add a success message

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -67,6 +70,24 @@ const mutations = {
 	},
 	[types.SET_TAXONOMY_TERM_EDITOR_MODE] (state, payload) {
 		set(state, 'editorMode', payload);
+	},
+	[types.REORDER_TERMS] (state, list) {
+		const updatedList = list.map((item, index) => {
+			return {
+				...item,
+				orderNumber: index
```
kvas-damian: This will break after changing the parent of a term. Let's use `orderNumber` instead.

bognix: Let's first sort by order number, then use index. This will fix possible holes after implementing delete.

resources/assets/js/admin/components/taxonomies/TaxonomyTermItem.vue
rogatty: We should be consistent with other draggable components but I like having a drag handle marked by an icon. If the whole item is draggable then it's easy for users to make a mistake.

resources/assets/js/admin/store/modules/taxonomyTerms.js
bognix: Current implementation displays all alert messages so success messages after using arrows look very ugly. I pushed the change anyway so you can check it out.

Maybe we can change the implementation and only show the last dismissable alert.

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -67,6 +72,24 @@ const mutations = {
...
+	},
+	[types.REORDER_TERMS] (state, list) {
+		const updatedList = list
+			.sort((termA, termB) => termA - termB)
+			.map((item, index) => ({
+				...item,
+				orderNumber: index
+			}));
+
+		set(state, 'terms', state.terms.map(stateTerm => {
```
rogatty: We're replacing the whole list here, smells like Redux 😉 I don't have anything against it, but @kvas-damian mentioned it's not a good practice in Vuex.

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -83,8 +106,24 @@ const actions = {
...
+					orderNumbers[term.parent_id] = orderNumbers[term.parent_id] + 1;
+				}
+				term.orderNumber = orderNumbers[term.parent_id];
+
+				return term;
+			});
+
+			commit(
+				types.SETUP_TERMS,
+				termsWithOrderNumbers.map(term => includeTag(term, included.tags), termsWithOrderNumbers)
```
rogatty:
```suggestion
				termsWithOrderNumbers.map(term => includeTag(term, included.tags))
```

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -97,27 +136,65 @@ const actions = {
 		try {
 			const response = await axios.post(getApiUrl('taxonomy_terms?include=tags'), taxonomyTerm);
 			const {data: {included, ...term}} = response;
-			commit(types.ADD_TERM, includeTag(term, included.tags));
+			commit(types.ADD_TERM, includeTag(term, included.tags), state.terms);
```
rogatty:
```suggestion
			commit(types.ADD_TERM, includeTag(term, included.tags));
```

resources/assets/js/admin/store/modules/taxonomyTerms.js
rogatty: On the other hand, if we're returning the same object from the `map` callback, is it still a replacement? 🤔

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -97,27 +136,65 @@ const actions = {
...
-			commit(types.UPDATE_TERM, includeTag(term, included.tags), state.terms);
+
+			const {parent_id: oldParentId} = getters.termById(taxonomyTerm.id);
+			const {parent_id: updatedParentId} = taxonomyTerm;
+
+			if (oldParentId !== updatedParentId) {
+				taxonomyTerm.orderNumber = getters.getChildrenByParentId(updatedParentId).length;
+			}
+
+			commit(types.UPDATE_TERM, includeTag({...taxonomyTerm, ...term}, included.tags), state.terms);
```
rogatty: `taxonomyTerm` and `term` are really unhelpful names. How about using the occasion to name them better?

rogatty:
```suggestion
			commit(types.UPDATE_TERM, includeTag({...taxonomyTerm, ...term}, included.tags));
```

resources/assets/js/admin/store/modules/taxonomyTerms.js
rogatty: Let's not change the alert implementation, there are places like tag deletion where we want to display all alerts, not just the last. Also, I don't think it's a good UX.

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -97,27 +136,65 @@ const actions = {
...
 		} catch (error) {
 			throw error;
 		} finally {
 			commit(types.SET_TAXONOMY_TERMS_SAVING, false);
 		}
 	},
 
+	async moveTerm({commit, dispatch}, {terms, oldIndex, newIndex}) {
+		const direction = newIndex - oldIndex;
```
rogatty: It's a bit odd that we first convert direction to index in https://github.com/bethinkpl/wnl-platform/pull/1229/files#diff-266646e1c3c91981151eb930a8256a43R40 and then revert it here. Is it simpler than calculating the indexes here?

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -97,27 +136,65 @@ const actions = {
...
+		terms.splice(newIndex, 0, term);
+
+		commit(types.REORDER_TERMS, terms);
+
+		try {
+			await axios.put(getApiUrl('taxonomy_terms/move'), {
+				term_id: term.id,
+				direction
+			});
+			dispatch('addAutoDismissableAlert', {
```
rogatty: I think we should keep the UI logic inside the components. IMO, the data store shouldn't control alerts.

resources/assets/js/admin/components/taxonomies/TaxonomyTermEditorForm.vue
```diff
@@ -79,7 +80,7 @@ export default {
 	},
 	computed: {
 		...mapState('taxonomyTerms', ['terms', 'isSaving']),
-		...mapGetters('taxonomyTerms', ['getAncestorsById']),
+		...mapGetters('taxonomyTerms', ['getAncestorsById', 'getSiblingsByParentId']),
```
kvas-damian: `getSiblingsByParentId` isn't used now

resources/assets/js/admin/components/taxonomies/TaxonomyTermEditorForm.vue
```diff
@@ -71,6 +71,7 @@ export default {
 			id: null,
 			tag: null,
 			parent: null,
+			orderNumber: null
```
kvas-damian: unused

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -67,6 +72,24 @@ const mutations = {
 	},
 	[types.SET_TAXONOMY_TERM_EDITOR_MODE] (state, payload) {
 		set(state, 'editorMode', payload);
+	},
+	[types.REORDER_TERMS] (state, list) {
+		const updatedList = list
+			.sort((termA, termB) => termA - termB)
```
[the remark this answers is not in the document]
kvas-damian: Does this sort handle `undefined` that we introduce when we add a new term?

resources/assets/js/admin/store/modules/taxonomyTerms.js
kvas-damian: Yes, according to guidelines we should be updating selected items by `splice()` or `set()`: https://vuejs.org/v2/guide/list.html#Caveats

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -97,27 +136,65 @@ const actions = {
...
+	async moveTerm({commit, dispatch}, {terms, oldIndex, newIndex}) {
+		const direction = newIndex - oldIndex;
+
+		if (direction === 0) return;
+
+		const term = terms[oldIndex];
+		terms.splice(oldIndex, 1);
+		terms.splice(newIndex, 0, term);
+
+		commit(types.REORDER_TERMS, terms);
```
kvas-damian: I think this operation should be performed only after success response from the server.
In case of an error, we'll present the inconsistent state. Additionally, in the UI there is no way to resubmit the operation.

resources/assets/js/admin/store/modules/taxonomyTerms.js
bognix: It won't break because we are calling "set", without it it would break.

resources/assets/js/admin/store/modules/taxonomyTerms.js
bognix: But this look odd that you move things, nothing happens (because request is pending) and then boom! suddenly the UI updates. We can revert it on failure but I wouldn't wait with updating the UI.

kvas-damian: great idea

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -97,27 +136,65 @@ const actions = {
...
+
+			commit(types.UPDATE_TERM, includeTag({...taxonomyTerm, ...term}, included.tags), state.terms);
 		} catch (error) {
 			throw error;
 		} finally {
 			commit(types.SET_TAXONOMY_TERMS_SAVING, false);
 		}
 	},
 
+	async moveTerm({commit, dispatch}, {terms, oldIndex, newIndex}) {
```
kvas-damian: I remember us talking about blocking the UI during the order update to make sure that previous request is finished before performing a new one (to make sure the requests are received by the server in the right order). I can't find it in the code...

resources/assets/js/admin/store/modules/taxonomyTerms.js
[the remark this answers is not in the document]
bognix: yes it was :D

resources/assets/js/admin/store/modules/taxonomyTerms.js
bognix: what is more, if the `updatedList` has a lot of items then making bunch of splices and sets may have bad impact on performance - it's the same as making bunch of `sets` - the same thing we experience in `qna.js` recently.

resources/assets/js/admin/store/modules/taxonomyTerms.js
bognix: any suggestions? `taxonomyTerm` and `updatedTaxonomyTerm` are OK?

resources/assets/js/admin/store/modules/taxonomyTerms.js
bognix: because it's not there. I'm not convinced about this. It will really decrease the User Experience and will be removed after we implement batch update / publish changes functionality.

resources/assets/js/admin/store/modules/taxonomyTerms.js
bognix: This method has consistent API and as I commented above next to the line you mentioned, the purpose of having those two methods is to unify the input and pass it to the `moveTerm` method.

resources/assets/js/admin/store/modules/taxonomyTerms.js
rogatty: Let's keep it consistent with `if (oldParentId !== updatedParentId) {`

resources/assets/js/admin/store/modules/taxonomyTerms.js
bognix: Change of approach, the list has to be in right order and the orderNumbers are added based on the item's order in the list. I added getter that returns the list in sorted order and it's the source of the passed list here.

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -83,41 +106,90 @@ const actions = {
...
 		commit(types.SET_TAXONOMY_TERMS_SAVING, true);
 		try {
-			const response = await axios.put(getApiUrl(`taxonomy_terms/${taxonomyTerm.id}?include=tags`), taxonomyTerm);
-			const {data: {included, ...term}} = response;
-			commit(types.UPDATE_TERM, includeTag(term, included.tags), state.terms);
+			const response = await axios.put(getApiUrl(`taxonomy_terms/${originalTerm.id}?include=tags`), originalTerm);
+			const {data: {included, ...updatedTerm}} = response;
+
+			const {parent_id: originalParentId} = getters.termById(originalTerm.id);
+			const {parent_id: updatedParentId} = originalTerm;
```
kvas-damian:
```suggestion
			const {parent_id: updatedParentId} = updatedTerm;
```

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -131,68 +131,63 @@ const actions = {
...
+		allSiblings.splice(oldIndex, 1);
+		allSiblings.splice(newIndex, 0, term);
 
-		const term = terms[oldIndex];
-		terms.splice(oldIndex, 1);
-		terms.splice(newIndex, 0, term);
+		commit(types.TERMS_COUNT_ORDER_NUMBERS, {list: allSiblings});
 
-		commit(types.REORDER_TERMS, terms);
+		return allSiblings;
```
rogatty: No need to return as we're not using it anywhere.

bognix: true, the plan was different :D

resources/assets/js/admin/components/taxonomies/TaxonomyTermItem.vue
```diff
@@ -89,6 +89,10 @@ export default {
 			type: Object,
 			required: true,
 		},
+		loading: {
```
rogatty: `isLoading`?

resources/assets/js/admin/components/taxonomies/TaxonomyTermsList.vue
```diff
@@ -4,15 +4,25 @@
 			<wnl-taxonomy-term-item
 				v-for="term in terms"
 				:term="term"
+				:loading="isSaving"
```
rogatty: How about using `isSaving` directly inside `wnl-taxonomy-term-item`?

resources/assets/js/admin/store/modules/taxonomyTerms.js
```diff
@@ -169,25 +169,38 @@ const actions = {
...
+		} catch (e) {
+			commit(types.SET_TAXONOMY_TERMS_SAVING, false);
+			return;
+		}
 
-		await axios.put(getApiUrl('taxonomy_terms/move'), {
-			term_id: term.id,
-			direction
-		});
+		try {
```
rogatty: A single `try` block should be sufficient.

PR #1232, opened 2019-01-17
PLAT-921 | Handle multi-line headers in slides

rogatty:
https://bethink.atlassian.net/browse/PLAT-921
Fix styles and slide snippet header.

### DIFF
--- a/lib/SlideParser/Parser.php
+++ b/lib/SlideParser/Parser.php
@@ -27,7 +27,7 @@ class Parser
 
 	const LUCID_EMBED_PATTERN = '/<div[^\<]*<iframe.*lucidchart.com\/documents\/embeddedchart\/([^"]*).*<\/iframe>[^\<]*<\/div>/';
 
-	const HEADER_PATTERN = '/<h.*>([\s\S]*)<\/h.*>/';
+	const HEADER_PATTERN = '/<h[1-6]>([\s\S]*)<\/h[1-6]>/';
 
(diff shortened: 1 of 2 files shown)


lib/SlideParser/Parser.php
```diff
@@ -27,7 +27,7 @@ class Parser
 
 	const LUCID_EMBED_PATTERN = '/<div[^\<]*<iframe.*lucidchart.com\/documents\/embeddedchart\/([^"]*).*<\/iframe>[^\<]*<\/div>/';
 
-	const HEADER_PATTERN = '/<h.*>([\s\S]*)<\/h.*>/';
+	const HEADER_PATTERN = '/<h[1-6].*>([\s\S]*)<\/h[1-6]>/U';
```
kvas-damian: Maybe that won't happen, but I'm a little bit worried about whitespaces in the closing tag.

Btw you're using the greedy selector, so it can happen that we mix two headers. This is a famous problem in the code we all know: https://github.com/wikimedia/mediawiki/blob/master/includes/parser/Parser.php#L1678

lib/SlideParser/Parser.php
```diff
@@ -403,7 +403,11 @@ public function createSnippet($slideHtml)
 		$match = $this->match(self::HEADER_PATTERN, $slideHtml);
 
 		if ($match) {
-			$snippet['header'] = strip_tags($match[0][1]);
+			$header = $match[0][1];
+			$header = str_replace('<br>', ' ', $header);
```
kvas-damian: what if we get `<br />`?

PR #1235, opened 2019-01-21
Enable vue/no-parse-error eslint rule

rogatty:


resources/assets/js/components/quiz/QuizList.vue
```diff
@@ -24,13 +24,13 @@
...
-		<p class="has-text-centered" v-if="!plainList && !displayResults">
+		<p v-if="!plainList && !displayResults" class="has-text-centered">
 			<a class="button is-primary" :class="{'is-loading': isProcessing}" @click="verify">
 				Sprawdź wyniki
 			</a>
-			<p v-if="canEndQuiz && !displayResults" class="has-text-centered margin vertical">
-				<a class="link" @click="$emit('checkQuiz', true)">Przerwij test i sprawdź wyniki</a>
-			</p>
+		</p>
+		<p v-if="!plainList && canEndQuiz && !displayResults" class="has-text-centered margin vertical">
```
rogatty: `p` in `p` is invalid so I moved this one out and added `!plainList` to the condition so the logic is exactly the same as before.

PR #1238, opened 2019-01-22
[PLAT-906] Nested course structure

gitkuba:


app/Http/Controllers/Api/PrivateApi/CourseStructureNodesApiController.php
```diff
@@ -0,0 +1,24 @@
...
+
+	public function __construct(Request $request)
+	{
+		parent::__construct($request);
+		$this->resourceName = config('papi.resources.course-structure-nodes');
+	}
+
+	public function getByCourseId($courseId)
+	{
+		$nodes = CourseStructureNode::where('course_id', $courseId)->get()->toFlatTree();
```
kvas-damian:
```suggestion
		$nodes = CourseStructureNode::where('course_id', $courseId)->defaultOrder()->get()->toFlatTree();
```

app/Http/Controllers/Api/Transformers/CourseStructureNodeTransformer.php
```diff
@@ -0,0 +1,55 @@
...
+		if ($this->parent) {
+			$data = array_merge($data, $this->parent);
+		}
+
+		return $data;
+	}
+
+	public function includeLessons(CourseStructureNode $node)
+	{
+		if ($node->structurable_type === 'App\\Models\\Lesson') {
```
kvas-damian: use `App\\Models\\Lesson:class` instead of hardcoding a string

database/migrations/2019_01_18_151943_create_course_structure_nodes_table.php
```diff
@@ -0,0 +1,36 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('course_structure_nodes', function (Blueprint $table) {
+	        $table->increments('id');
+	        $table->unsignedInteger('course_id');
+	        // TODO: discussion trigger - add here course_id ? Or maybe structure_id, as there may be many structures within a course?
```
kvas-damian: I don't know all use-cases. We can ask @adamkarminski if he identifies any use-cases including multiple structures.
Without any additional knowledge, I'd say that course have only one structure.

resources/assets/js/admin/store/mutations-types.js
```diff
@@ -36,3 +36,6 @@ export const SET_EXPANDED_TAXONOMY_TERMS = 'EXPAND_TAXONOMY_TERMS';
 export const SET_TAGS_LOADING = 'SET_TAGS_LOADING';
 export const SETUP_TAGS = 'SETUP_TAGS';
 export const ADD_TAG = 'ADD_TAG';
+
+// Structure nodes
+export const SETUP_NODES = 'SETUP_NODES';
```
kvas-damian:
```suggestion
export const SETUP_COURSE_STRUCTURE = 'SETUP_COURSE_STRUCTURE';
```

kvas-damian: maybe `SETUP_COURSE_STRUCTURE`?

I'm also considering if we should name it `courseStructureNodes` or simply `courseStructure`. The node would be reserved for model only.

database/migrations/2019_01_18_151943_create_course_structure_nodes_table.php
gitkuba: Collections is a possible use case i think.

resources/assets/js/admin/components/structures/StructureNode.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+	<div>
+		<!--<div>{{ node.structurable_type }} {{ node.structurable_id }}</div>-->
```
kvas-damian:
```suggestion
```

resources/assets/js/admin/store/mutations-types.js
gitkuba: Depends on what we decide regarding multiple structures within a course.

resources/assets/js/admin/components/structures/StructureEditor.vue
```diff
@@ -0,0 +1,50 @@
...
+<script>
+	import StructureNode from 'js/admin/components/structures/StructureNode';
+	import {mapActions, mapState} from 'vuex';
+	import {getApiUrl} from 'js/utils/env';
+
+	export default {
+		name: 'StructureEditor',
+		props: {
+			courseId: {
+				required: true,
```
kvas-damian: type?

resources/assets/js/admin/components/structures/StructureEditor.vue
```diff
@@ -0,0 +1,50 @@
...
+			}
+		},
+		computed: {
+			rootNodes() {
+				return this.nodes.filter(node => node.parent_id === null);
```
kvas-damian: you can define getter on a store.
Even better, if you could add an abstraction layer over the TaxonomyTerms and reuse that code.

app/Http/Controllers/Api/Transformers/CourseStructureNodeTransformer.php
```diff
@@ -0,0 +1,56 @@
...
+		$this->parent = $parent;
+	}
+
+	public function transform(CourseStructureNode $node)
+	{
+		$data = [
+			'id'                => $node->id,
+			'structurable_type' => $node->structurable_type,
+			'structurable_id'   => $node->structurable_id,
+			'order_number'      => $node->order_number,
```
kvas-damian:
```suggestion
```

resources/assets/js/admin/store/modules/structureNodes.js
```diff
@@ -0,0 +1,54 @@
...
+// Mutations
+const mutations = {
+	[types.SETUP_NODES](state, payload) {
+		set(state, 'nodes', payload);
+	},
+};
+
+// Actions
+const actions = {
+	async fetchStructure({commit}, courseId) {
```
kvas-damian: you're fetching a structure, but setting nodes. Maybe we could unify that?

resources/assets/js/admin/components/structures/StructureNode.vue
```diff
@@ -0,0 +1,65 @@
...
+</style>
+
+<script>
+	import StructureNode from 'js/admin/components/structures/StructureNode';
+	import GroupNodeElement from 'js/admin/components/structures/GroupNodeElement';
+	import LessonNodeElement from 'js/admin/components/structures/LessonNodeElement';
+	import {mapState} from 'vuex';
+
+	export default {
+		name: "StructureNode",
```
kvas-damian:
```suggestion
```

resources/assets/js/admin/components/structures/GroupNodeElement.vue
```diff
@@ -0,0 +1,21 @@
...
+<style lang="sass" rel="stylesheet/sass">
+
+</style>
+
+<script>
+	export default {
+		name: 'GroupNodeElement',
+		props: ['structurable'],
+		data(){
+			return {...this.structurable}
```
kvas-damian: Can you be more explicit about which fields we use? Or maybe we should use `structurable` directly?

resources/assets/js/admin/components/structures/GroupNodeElement.vue
```diff
@@ -0,0 +1,21 @@
...
+</template>
+
+<style lang="sass" rel="stylesheet/sass">
+
+</style>
+
+<script>
+	export default {
+		name: 'GroupNodeElement',
+		props: ['structurable'],
```
kvas-damian: props validation

resources/assets/js/admin/components/structures/GroupNodeElement.vue
```diff
@@ -0,0 +1,21 @@
...
+<style lang="sass" rel="stylesheet/sass">
+
+</style>
+
+<script>
+export default {
+	name: 'GroupNodeElement',
+	props: ['structurable'],
+	data(){
+		return {...this.structurable};
```
kvas-damian: Can you be more explicit about which fields we use? Or maybe we should use `structurable` directly?

resources/assets/js/admin/components/structures/StructureNode.vue
```diff
@@ -0,0 +1,65 @@
...
+</template>
+
+<style lang="sass" scoped>
+	@import 'resources/assets/sass/variables'
+	@import 'resources/assets/sass/mixins'
+
+	.structure-node
+
+		.child-node
+			margin-left: $margin-big
```
kvas-damian: this style should be defined inside `structure-node`

resources/assets/js/admin/store/modules/structureNodes.js
```diff
@@ -0,0 +1,54 @@
...
+	isLoading: false,
+	nodes: [],
+};
+
+// Getters
+const getters = {};
+
+// Mutations
+const mutations = {
+	[types.SETUP_NODES](state, payload) {
```
kvas-damian: this mutation name can easily collapse with other stores mutation names. Can we be more explicit?

resources/assets/js/admin/store/modules/structureNodes.js
```diff
@@ -0,0 +1,54 @@
...
+
+// Actions
+const actions = {
+	async fetchStructure({commit}, courseId) {
+		try {
+			const response = await axios.get(getApiUrl(`course_structure_nodes/${courseId}?include=lessons,groups`));
+			const {data: {included, ...nodes}} = response;
+			commit(types.SETUP_NODES, Object.values(nodes).map(node => _parseIncludes(node, included)));
+		} catch (error) {
+			$wnl.logger.capture(error);
```
kvas-damian: error message?

app/Console/Commands/MigrateCourseStructure.php
```diff
@@ -0,0 +1,79 @@
...
+use Illuminate\Console\Command;
+
+class MigrateCourseStructure extends Command
+{
+	/**
+	 * The name and signature of the console command.
+	 *
+	 * @var string
+	 */
+	protected $signature = 'course:migrate-structure';
```
bognix: Some time ago I had a discussion with @kvas-damian how to distinguish between commands that are meant for "one time" migrations and which should live longer. Here are some conclusions we made:
* Create a command in namespace "data-migration". The example for this is the `MigrateQna` command which has signature: `protected $signature = 'data-migration:attach-qna-to-discussion';`
* Execute data migrations as part of database migrations to make sure the data is migrated after release and we won't forget about it. What is more it adds extra context to command origins :)

app/Console/Commands/MigrateCourseStructure.php
```diff
@@ -0,0 +1,79 @@
...
+	 *
+	 * @return mixed
+	 */
+	public function handle()
+	{
+		$groups = Group::all();
+
+		foreach ($groups as $group) {
+			$groupNode = CourseStructureNode::create([
+				'course_id'         => 1,
```
bognix: image

app/Console/Commands/MigrateCourseStructure.php
```diff
@@ -0,0 +1,79 @@
...
+				'structurable_id'   => $group->id,
+			]);
+			foreach ($group->lessons as $lesson) {
+				CourseStructureNode::create([
+					'type'              => 'lesson',
+					'course_id'         => 1,
+					'structurable_type' => 'App\\Models\\Lesson',
+					'structurable_id'   => $lesson->id,
+				], $groupNode);
+				print 'L';
```
bognix: do you want it to stay in the code? :)

app/Console/Commands/MigrateCourseStructure.php
```diff
@@ -0,0 +1,79 @@
...
+
+		foreach ($groups as $group) {
+			$groupNode = CourseStructureNode::create([
+				'course_id'         => 1,
+				'structurable_type' => 'App\\Models\\Group',
+				'structurable_id'   => $group->id,
+			]);
+			foreach ($group->lessons as $lesson) {
+				CourseStructureNode::create([
+					'type'              => 'lesson',
```
bognix: why we don't add type to group as well?

app/Http/Controllers/Api/Transformers/CourseStructureNodeTransformer.php
```diff
@@ -0,0 +1,55 @@
...
+		if ($this->parent) {
+			$data = array_merge($data, $this->parent);
+		}
+
+		return $data;
+	}
+
+	public function includeLessons(CourseStructureNode $node)
+	{
+		if ($node->structurable_type === Lesson::class) {
```
bognix: We have a method in ApiController called `protected static function getResourceTransformer($resource)` and it can fetch the transformer for given resource. Maybe we can reuse this method? It's static so it shouldn't be a problem. Maybe this method should be placed somewhere else, that I don't know - it's up to you.

Using some method to get the transformer for the node won't require any code changes in the future if we decide to support different structurable types

resources/assets/js/admin/components/structures/GroupNodeElement.vue
```diff
@@ -0,0 +1,21 @@
...
+
+<style lang="sass" rel="stylesheet/sass">
+
+</style>
+
+<script>
+export default {
+	name: 'GroupNodeElement',
+	props: {
+		structurable :{
```
bognix:
```suggestion
		structurable: {
```

resources/assets/js/admin/components/structures/GroupNodeElement.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+	<div class="margin bottom">
+		{{ structurable.name }}
+	</div>
+</template>
+
+<style lang="sass" rel="stylesheet/sass">
```
bognix: it's redundant.

resources/assets/js/admin/components/structures/StructureNode.vue
```diff
@@ -0,0 +1,62 @@
+<template>
+	<div class="structure-node">
+		<component :is="structurableComponent" :structurable="node.structurable"></component>
+		<structure-node
+				v-for="childNode in childNodes"
+				:key="childNode.id"
```
bognix: id doesn't guarantee the keys will be different, group and lesson can have the same id.

resources/assets/js/admin/components/structures/StructureNode.vue
```diff
@@ -0,0 +1,62 @@
+<template>
+	<div class="structure-node">
+		<component :is="structurableComponent" :structurable="node.structurable"></component>
```
bognix: I'm not sure if I would break Lesson and Group component into two separate components at this point. They are the same right now after all.

resources/assets/js/admin/components/structures/StructureEditor.vue
```diff
@@ -0,0 +1,58 @@
+<template>
+	<div>
+		<div class="header">
+			<h2 class="title is-2">Edycja struktury kursu</h2>
+		</div>
+
+		<structure-node v-for="(node, index) in rootNodes" :key="index" :node="node"></structure-node>
```
bognix: each node have id, let's use id for key.

resources/assets/js/admin/components/structures/StructureEditor.vue
```diff
@@ -0,0 +1,58 @@
...
+
+	</div>
+</template>
+
+<style lang="sass" rel="stylesheet/sass" scoped>
```
bognix: redundandt.

resources/assets/js/admin/components/structures/StructureEditor.vue
```diff
@@ -0,0 +1,58 @@
...
+	@import 'resources/assets/sass/variables'
+	@import 'resources/assets/sass/mixins'
+
+</style>
+
+<script>
+import StructureNode from 'js/admin/components/structures/StructureNode';
+import {mapActions, mapState} from 'vuex';
+import {getApiUrl} from 'js/utils/env';
```
bognix: this is not used.

resources/assets/js/admin/components/structures/StructureEditor.vue
```diff
@@ -0,0 +1,58 @@
...
+		courseId: {
+			required: true,
+			type: String,
+		}
+	},
+	computed: {
+		rootNodes() {
+			return this.nodes.filter(node => node.parent_id === null);
+		},
+		...mapState('courseStructure', {
```
bognix: `...mapState('courseStructure', ['nodes']`

no need to use an object if you don't change the name.

app/Console/Commands/MigrateCourseStructure.php
gitkuba: yes!

app/Http/Controllers/Api/Transformers/CourseStructureNodeTransformer.php
gitkuba: We'll still need to create new "include" method for each new "structurable". Moreover it's required to return from such method only when "structurable" is of a certain type.

resources/assets/js/admin/components/structures/StructureNode.vue
gitkuba: I'm not sure as well. But I have a hunch that styling and logic could be slightly different :)

resources/assets/js/admin/components/structures/StructureEditor.vue
bognix: @gitkuba it's great you resolved it, but could you fix it?

resources/assets/js/admin/components/structures/StructureEditor.vue
```diff
@@ -0,0 +1,54 @@
...
+		courseId: {
+			required: true,
+			type: Number|String,
+		}
+	},
+	computed: {
+		rootNodes() {
+			return this.nodes.filter(node => node.parent_id === null);
+		},
+		...mapState('courseStructure', ['nodes'])
```
bognix: could you ...mapState as first in computed list?

resources/assets/js/admin/components/structures/StructureNode.vue
```diff
@@ -0,0 +1,62 @@
...
+
+export default {
+	name: 'StructureNode',
+	props: {
+		node: {
+			type: Object,
+			required: true,
+		}
+	},
+	components: {
```
bognix: could you move components below the name in order to keep things consistent across components.

app/Console/Commands/MigrateCourseStructure.php
```diff
@@ -0,0 +1,78 @@
...
+					'course_id'         => 1,
+					'structurable_type' => 'App\\Models\\Lesson',
+					'structurable_id'   => $lesson->id,
+				], $groupNode);
+				print 'L';
+			}
+			print 'G';
+		}
+
+		// TODO: Below code is for testing purposes. To be removed.
```
rogatty: When are we going to remove it?

resources/assets/js/admin/components/structures/LessonNodeElement.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+	<div class="margin bottom">
+		<router-link :to="to">{{ structurable.name }}</router-link>
+	</div>
+</template>
+
+<style lang="sass" rel="stylesheet/sass">
+
+</style>
```
rogatty: Can be removed.

resources/assets/js/admin/components/structures/StructureEditor.vue
```diff
@@ -0,0 +1,54 @@
...
+
+		<structure-node v-for="node in rootNodes" :key="node.id" :node="node"></structure-node>
+	</div>
+</template>
+
+<style lang="sass" rel="stylesheet/sass" scoped>
+	@import 'resources/assets/sass/variables'
+	@import 'resources/assets/sass/mixins'
+
+</style>
```
rogatty: Can be removed.

resources/assets/js/admin/components/structures/StructureEditor.vue
```diff
@@ -0,0 +1,54 @@
...
+
+<script>
+import StructureNode from 'js/admin/components/structures/StructureNode';
+import {mapActions, mapState} from 'vuex';
+
+export default {
+	name: 'StructureEditor',
```
rogatty: I believe it's not needed.

resources/assets/js/admin/components/structures/StructureNode.vue
rogatty: We can simplify during CRUD implementation if it's not needed.

resources/assets/js/admin/components/structures/StructureNode.vue
```diff
@@ -0,0 +1,62 @@
...
+				v-for="childNode in childNodes"
+				:key="childNode.id"
+				:node="childNode">
+		</structure-node>
+	</div>
+</template>
+
+<style lang="sass" scoped>
+	@import 'resources/assets/sass/variables'
+	@import 'resources/assets/sass/mixins'
```
rogatty: I don't think we're using any mixins here.

resources/assets/js/admin/components/structures/StructureNode.vue
```diff
@@ -0,0 +1,62 @@
...
+	},
+	components: {
+		StructureNode,
+		GroupNodeElement,
+		LessonNodeElement,
+	},
+	data() {
+		return {
+			typeComponentMap: {
+				'App\\Models\\Group': 'GroupNodeElement',
```
rogatty: You can use references instead of strings, see https://github.com/bethinkpl/wnl-platform/blob/63442f1eed4578ff2ff30d271e3806e15c582ccf/resources/assets/js/admin/components/taxonomies/TaxonomyTermEditorRight.vue#L49

resources/assets/js/admin/store/modules/courseStructure.js
```diff
@@ -0,0 +1,50 @@
...
+	async fetchStructure({commit}, courseId) {
+		const response = await axios.get(getApiUrl(`course_structure_nodes/${courseId}?include=lessons,groups`));
+		const {data: {included, ...nodes}} = response;
+		commit(types.SETUP_COURSE_STRUCTURE, Object.values(nodes).map(node => _parseIncludes(node, included)));
+	},
+};
+
+const _parseIncludes = (node, included) => {
+	if (node.hasOwnProperty('groups')) {
+		node.structurable = included.groups[node.groups[0]];
```
rogatty: Won't this break the reactivity? Shouldn't we use `Vue.set`? @bognix @kvas-damian?

rogatty: We'll probably reuse this method when updating nodes, but we can get back to the topic then.

resources/assets/js/admin/components/structures/StructureNode.vue
gitkuba: Very nice idea! Are you ok with doing that as part of CRUD?

PR #1239, opened 2019-01-23
Store Questions Bank progress in db

bognix:


app/Http/Controllers/Api/PrivateApi/UserStateApiController.php
```diff
@@ -105,25 +105,43 @@ public function incrementTime(Request $request, $user)
...
-	public function saveQuizPosition(Request $request, $user)
+	public function saveQuizPosition(Request $request)
 	{
-		$cacheKey = $this->hashedFilters($request->filters);
-		$cacheTags = $this->getFiltersCacheTags(config('papi.resources.quiz-questions'), $user);
-		Cache::tags($cacheTags)->put($cacheKey, $request->position, 60 * 24);
+		$key = $this->hashedFilters($request->filters);
+		$value = $request->position;
+		$state = UserQuestionsBankState::firstOrNew(
+			['user_id' => $request->route('id')]
```
gitkuba: How about some kind of validation? Just to make sure we won't try to save `0` or `undefined` as id?

app/Http/Controllers/Api/PrivateApi/UserStateApiController.php
```diff
@@ -105,25 +105,43 @@ public function incrementTime(Request $request, $user)
...
+		if (!Auth::user()->can('update', $state)) {
+			return $this->respondForbidden();
+		}
+
+
+		$state->key = $key;
+		$state->value = json_encode($value);
+		$state->save();
 
 		return $this->json([
```
gitkuba: Can we use `respondOk` for consistency?

app/Http/Controllers/Api/PrivateApi/UserStateApiController.php
```diff
@@ -105,25 +105,43 @@ public function incrementTime(Request $request, $user)
...
+			['user_id' => $request->route('id')]
+		);
+
+		if (!Auth::user()->can('update', $state)) {
+			return $this->respondForbidden();
+		}
+
+
+		$state->key = $key;
+		$state->value = json_encode($value);
```
gitkuba: How about using json <-> array casting in model?

bognix: cast only works one way - on read, not both ways.

app/Http/Controllers/Api/PrivateApi/UserStateApiController.php
```diff
@@ -105,25 +105,42 @@ public function incrementTime(Request $request, $user)
...
 	{
-		$cacheKey = $this->hashedFilters($request->filters);
-		$cacheTags = $this->getFiltersCacheTags(config('papi.resources.quiz-questions'), $user);
-		$cachedPosition = Cache::tags($cacheTags)->get($cacheKey, $request->position);
+		$key = $this->hashedFilters($request->filters);
+
+		$state = UserQuestionsBankState::select(['value', 'user_id'])
+			->where('user_id', $request->route('id'))
+			->where('key', $key)
+			->first();
```
rogatty: Shouldn't this be a model method?

app/Http/Controllers/Api/PrivateApi/UserStateApiController.php
```diff
@@ -105,25 +105,42 @@ public function incrementTime(Request $request, $user)
...
+
+		if (!empty($state) && !Auth::user()->can('view', $state)) {
```
rogatty: What is `!empty($state)` for? Do we want to allow other users to know that someone has an empty state?

app/Http/Controllers/Api/PrivateApi/UserStateApiController.php
```diff
@@ -105,25 +105,42 @@ public function incrementTime(Request $request, $user)
...
+			return $this->respondForbidden();
+		}
 
 		return $this->json([
-			'position' => $cachedPosition,
+			'position' => json_decode($state->value ?? '')
```
rogatty: Shouldn't this be an empty array or object?

PR #1245, opened 2019-01-24
PLAT-934 | Notify Prometheus Pushgateway after scheduled job is run

rogatty:
See https://pushgateway.bethink.tech/ for an example how it looks there.

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -33,7 +33,7 @@ class Kernel extends ConsoleKernel
 			->hourly()
 			->withoutOverlapping()
 			->after(function () {
-				$this->notifyPrometheusPushGateway('orders_stats_export');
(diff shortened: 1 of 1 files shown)


app/Console/Kernel.php
```diff
@@ -39,43 +43,95 @@ protected function schedule(Schedule $schedule)
...
 		$schedule
 			->command('data-integrity:check')
-			->dailyAt('04:00');
+			->dailyAt('04:00')
+			->after(function () {
+				$this->notifyPrometheusPushGateway('data_integrity_check');
+			});
+	}
+
+	private function notifyPrometheusPushGateway($metricName)
```
kvas-damian: I'm not sure if this method should be a part of `Kernel.php` Maybe we should have a separate class for that?

rogatty: No one has a good idea where to put it so let's leave it as it is.

PR #1249, opened 2019-01-25
Enable eslint rules related to computed properties

rogatty:


resources/assets/js/components/qna/QnaQuestion.vue
```diff
@@ -266,15 +257,7 @@ export default {
 			return this.question.text;
 		},
 		author() {
-			if (this.question.hasOwnProperty('profiles')) {
```
rogatty: I found no scenario where this would be false, so I removed all the related code and fixed:
```
/resources/assets/js/components/qna/QnaQuestion.vue
  268:9  error  Expected to return a value in "author" computed property      vue/return-in-computed-property
  272:5  error  Unexpected side effect in "author" computed property          vue/no-side-effects-in-computed-properties
  273:5  error  Unexpected asynchronous action in "author" computed property  vue/no-async-in-computed-properties
```

It was introduced in https://github.com/bethinkpl/wnl-platform/pull/193/commits/4bfc273db7a3452fb086aa238bc29ce2e99744d2 and seems to no longer be valid.

resources/assets/js/components/user/plan/ManualPlan.vue
```diff
@@ -256,11 +256,13 @@ export default {
 			});
 		},
 		sortedManualStartDates() {
-			return this.manualStartDates.sort((a, b) => {
-				const dateA = new Date(a.startDate);
-				const dateB = new Date(b.startDate);
-				return dateA - dateB;
-			});
+			return this.manualStartDates
+				.slice()
```
kvas-damian: maybe instead of `.slice()` we could use `{...this.manualStartDates}`? I think it's better describing what we're trying to achieve.

rogatty: You mean `[...this.manualStartDates]`? I used that first but:
1. PHPStorm complains about types
2. `.slice()` is ten times faster, see https://stackoverflow.com/questions/3978492/fastest-way-to-duplicate-an-array-in-javascript-slice-vs-for-loop

PR #1250, opened 2019-01-28
PLAT-874 | Double comments

bognix:
When calling _.uniq make sure we are comparing the same types

After adding a comment to a slide with ID `x` we do following operations:
- we add it to comments/comments state
- we push it to slideshow/slides[x]/comments array

The problem was that newly created comment had an ID stored as an integer and the old comments have ids stored as strings in slideshow/slides[x]/comments array

When calling _.uniq method I want to make sure we are compering the same types.

Alternative approach would be to don't concat the old commentable comments with newly fetched ones after slide change. However this method is widely used in many modules and I'm afraid of side effects.

### DIFF
--- a/resources/assets/js/store/modules/comments.js
+++ b/resources/assets/js/store/modules/comments.js
@@ -136,10 +136,9 @@ export const commentsMutations = {
 		Object.keys(commentsResourceObj).forEach(resource => {
 			Object.keys(commentsResourceObj[resource]).forEach(resourceId => {
 				const oldComments = state[resource][resourceId].comments || [];
-				const commentsIds = _.uniq(
-					oldComments
-						.concat(Object.keys(commentsResourceObj[resource][resourceId]))
-						.map(id => id.toString())
+				const commentsIds = _.uniqBy(
+					oldComments.concat(Object.keys(commentsResourceObj[resource][resourceId])),
+					commentId => commentId.toString()
 				);
 				set(state[resource][resourceId], 'comments', commentsIds);
 			});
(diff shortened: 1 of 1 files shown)


resources/assets/js/store/modules/comments.js
```diff
@@ -136,7 +136,11 @@ export const commentsMutations = {
 		Object.keys(commentsResourceObj).forEach(resource => {
 			Object.keys(commentsResourceObj[resource]).forEach(resourceId => {
 				const oldComments = state[resource][resourceId].comments || [];
-				const commentsIds = _.uniq(oldComments.concat(Object.keys(commentsResourceObj[resource][resourceId])));
+				const commentsIds = _.uniq(
+					oldComments
+						.concat(Object.keys(commentsResourceObj[resource][resourceId]))
```
bognix: Alternative approach would be to remove the concat method and override the value of commentable comments with newly fetched ones. It will work well for slides but this action is also used for qna answers and quiz questions. I'm a bit afraid of side effects however after going through code I couldn't find a place where it could break.

resources/assets/js/store/modules/comments.js
```diff
@@ -136,7 +136,11 @@ export const commentsMutations = {
...
+						.map(id => id.toString())
```
rogatty: You could also use [`_.uniqBy`](https://lodash.com/docs/4.17.11#uniqBy) or [`_.uniqWith`](https://lodash.com/docs/4.17.11#uniqWith).

resources/assets/js/store/modules/comments.js
rogatty: IMO the best approach would be to unify the types, but it's a larger project. For now the current solution seems fine to me.

bognix: Not sure if it's a bigger project, `Object.keys` returns strings and it's the only place where we set commentables comments. So we either change the types of Object.keys returned values to Number or we cast the new comment id to string where it's added.

PR #1261, opened 2019-02-04
PLAT-964 | Fix resetting slides order

kvas-damian:
When we were fetching slides for sections we were taking them in the default Mysql order which isn't the same as `order_number`. We use those screens later to update order in the presentation: https://github.com/bethinkpl/wnl-platform/blob/c27353f3fa225cce9e3ba99dd174f8622681b97d/app/Console/Commands/ReorderSections.php#L98


rogatty:
If you're in the context and know how this script works, it would be great if you could expand this vague description: https://github.com/bethinkpl/wnl-platform/blob/d2b2fd12bee2f0e65ca1f60aef4e190ff221dfd2/app/Console/Commands/ReorderSections.php#L25

PR #1263, opened 2019-02-04
PLAT-965 | fix command for calculating final lek

bognix:
It's not fixing the problem on production but at least gives us a way to fix it afterwards.

### DIFF
--- a/app/Console/Commands/ExamsResults.php
+++ b/app/Console/Commands/ExamsResults.php
@@ -55,7 +55,7 @@ class ExamsResults extends Command
 			$user = User::find($userId);
 
 			if (!empty($results)) {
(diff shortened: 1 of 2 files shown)


app/Console/Commands/ExamsResults.php
```diff
@@ -55,7 +55,7 @@ public function handle()
 			$user = User::find($userId);
 
 			if (!empty($results)) {
-				$this->saveInDB($userId, $results);
+				$this->saveInDB($examTagId, $userId, $results);
```
rogatty: 🙈

rogatty: Would be great to have a PHP linter that would catch it.

PR #1266, opened 2019-02-04
PLAT-896 | Restore position on exam

bognix:


resources/assets/js/components/questions/QuestionsList.vue
```diff
@@ -319,11 +322,15 @@ export default {
 							return reject();
 						}
 					})
-					.catch(e => reject());
+					.catch(() => {
+						localStorage.removeItem(`wnl-exam-state-${this.currentUserId}`);
```
kvas-damian: we use `store-engine` as an abstraction layer to unify stores api.

resources/assets/js/components/questions/QuestionsList.vue
```diff
@@ -319,11 +326,15 @@ export default {
 							return reject();
 						}
 					})
-					.catch(e => reject());
+					.catch(() => {
+						localStorage.removeItem(this.localStorageKey);
```
rogatty: We display a quiz end confirmation and if user cancels, then we remove the state from storage? Doesn't seem right, but maybe I don't understand the flow.

resources/assets/js/components/questions/QuestionsList.vue
```diff
@@ -378,11 +389,23 @@ export default {
...
-			payload.answer === this.getQuestion(payload.id).selectedAnswer
-				&& !this.testMode
-				? this.onVerify(payload.id) || (payload.position && this.savePosition({position: payload.position}))
-				: this.selectAnswer(payload);
+		onSelectAnswer(payload, useLocalStorage = true) {
+			if (payload.answer === this.getQuestion(payload.id).selectedAnswer && !this.testMode) {
+				this.onVerify(payload.id) || (payload.position && this.savePosition({position: payload.position}));
+			} else {
+				this.selectAnswer(payload);
+				this.testMode && useLocalStorage && this.persistStateInLocalStorage();
```
rogatty: We already check `this.testMode` in `this.testMode`

resources/assets/js/components/questions/QuestionsList.vue
```diff
@@ -378,11 +389,23 @@ export default {
...
+			}
+		},
+		persistStateInLocalStorage() {
+			this.testMode && localStorage.setItem(this.localStorageKey, JSON.stringify({
+				results: this.testQuestions.map(question => {
+					return {
```
rogatty: Fun fact, you can use `question => ({...})` to return object from an arrow function.

resources/assets/js/components/questions/QuestionsList.vue
```diff
@@ -378,11 +389,23 @@ export default {
...
+						question_id: question.id,
+						answer_id: question.answers[question.selectedAnswer] && question.answers[question.selectedAnswer].id
```
rogatty: How about filtering out the questions without answers?

resources/assets/js/components/questions/QuestionsList.vue
bognix: I know, but this function had so many shapes :D I'll update.

resources/assets/js/components/questions/QuestionsList.vue
```diff
@@ -589,6 +661,9 @@ export default {
 			next();
 		}
 	},
+	created() {
+		this.testMode = !!this.presetOptionsToPass.examMode;
```
rogatty: It's already set in multiple places. Why do we add another one?

resources/assets/js/components/questions/QuestionsList.vue
```diff
@@ -521,61 +544,110 @@ export default {
...
+							context: this.context.value
+						});
+					})
+					.then(() => this.fetchQuestionsReactions(this.getPage(1)))
+					.then(() => this.reactionsFetched = true)
+					.then(() => this.fetchQuestionData(this.currentQuestion.id));
+			});
+		},
+		async restoreExamState(persistedState) {
+			const questions = JSON.parse(persistedState).results;
```
rogatty: I think this will fail if there was no persisted state or if it's malformed.

rogatty: You can move it to `try` block and it should be fine.

resources/assets/js/components/questions/QuestionsList.vue
```diff
@@ -521,61 +544,110 @@ export default {
...
+			} finally {
+				this.switchOverlay(false, 'testBuilding');
+			}
+		}
+	},
+	async mounted() {
+		try {
+			await this.setupQuestions();
+			const stateInLocalStorage = localStorage.getItem(this.localStorageKey);
+			if (stateInLocalStorage) {
```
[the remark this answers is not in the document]
bognix: no because it's not set when you use "Sprawdź się" view.

resources/assets/js/components/questions/QuestionsList.vue
```diff
@@ -521,61 +544,110 @@ export default {
...
+					time: timeBaseOnQuestions(questions.length),
+					examMode: true,
+					examQuestions: questionsIds
+				};
+				this.switchOverlay(true, 'testBuilding', 'testBuilding');
+				this.resetTest();
+				this.testMode = true;
+				await this.fetchQuestionsByIds(questionsIds);
+
+				questions.forEach(response => {
```
rogatty: It's not a response, it's a question from the local storage.

resources/assets/js/components/questions/QuestionsList.vue
```diff
@@ -378,11 +389,23 @@ export default {
...
+		onSelectAnswer(payload, useLocalStorage = true) {
+			if (payload.answer === this.getQuestion(payload.id).selectedAnswer && !this.testMode) {
+				this.onVerify(payload.id) || (payload.position && this.savePosition({position: payload.position}));
+			} else {
+				this.selectAnswer(payload);
+				this.testMode && useLocalStorage && this.persistStateInLocalStorage();
+			}
+		},
+		persistStateInLocalStorage() {
+			this.testMode && localStorage.setItem(this.localStorageKey, JSON.stringify({
```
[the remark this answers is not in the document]
kvas-damian: I'm not sure.
"Remaining" will allow users to cheat - don't answer, refresh and time is reset
"start" will include time when user was offline.

resources/assets/js/components/questions/QuestionsList.vue
bognix: Because confirm means I want to continue the test not cancel.

resources/assets/js/components/questions/QuestionsList.vue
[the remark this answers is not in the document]
bognix: by remaining I mean store the remaining time after clicking select answer.

I don't think I follow what you mean.

resources/assets/js/components/questions/QuestionsList.vue
bognix: no, it response from localStorage. It's the results object, I'll rename the `questions` variable.

resources/assets/js/components/questions/QuestionsList.vue
```diff
@@ -314,16 +322,21 @@ export default {
...
+					.then((dismiss) => {
 						if (dismiss === 'cancel') {
+							examStateStore.remove(this.examStateStoreKey);
 							return reject();
 						}
-					})
-					.catch(e => reject());
+					}, () => resolve())
+					.catch(() => {
+						examStateStore.remove(this.examStateStoreKey);
```
rogatty: If I'm correct, this block is called when SweetAlert fails somehow or the state removal fails. If it's the former, I think we should keep the state intact. If it's the latter, an immediate retry won't help.

bognix: I refactored this one, it was not working correctly previously (without my interference). If you pass two arguments to `then` the first one is called on success and the second one is called on error. Here it was implemented the other way around. I changed it and made it unified with other use cases of SweetAlert

rogatty: Makes sense but what is the `catch()` block for?

resources/assets/js/components/questions/QuestionsList.vue
```diff
@@ -624,7 +636,7 @@ export default {
 					loadingText: 'mockExam',
 					sizesToChoose: [results.length],
 					testQuestionsCount: results.length,
-					time: timeBaseOnQuestions(results.length),
+					time: storedState.time || timeBaseOnQuestions(results.length),
```
rogatty: Can `storedState.time` be 0?

resources/assets/js/components/questions/QuestionsList.vue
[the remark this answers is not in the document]
rogatty: You have 20 minutes left, think about an answer for 19 minutes, refresh the page, choose the answer and you still have 20 minutes left. Anyway, you cheat yourself.

PR #1267, opened 2019-02-05
[PLAT-945] Prevent redirect when token is invalid

gitkuba:


resources/assets/js/plugins/axios.js
```diff
@@ -29,10 +42,20 @@ export default (Vue, {store, router}) => {
 					return store.commit(types.USERS_SET_ACCOUNT_SUSPENDED, true);
 				}
 			}
+
 			if (error.response && error.response.status === 401) {
 				window.location.replace('/login');
```
gitkuba: Send a warning to sentry to identify other (than token mismatch) issues causing redirections.

resources/assets/js/plugins/axios.js
```diff
@@ -29,10 +42,20 @@ export default (Vue, {store, router}) => {
...
 			}
 
+			if (error.config && error.response && error.response.status === 419) {
```
gitkuba: Need to add some check to prevent infinite loop, in case `getCsrfToken()` results in `419`. Any good ideas?

config/session.php
```diff
@@ -29,7 +29,7 @@
     |
     */
 
-	'lifetime' => 120,
+	'lifetime' => 60 * 24,
```
gitkuba: I've extended session lifetime to lower the risk of redirecting to login. Of course long term solution is to show login modal. It shall be addressed as part of PLAT-967.

resources/assets/js/plugins/axios.js
```diff
@@ -1,7 +1,20 @@
 import axios from 'axios';
-import { get } from 'lodash';
+import {get} from 'lodash';
+import {getApiUrl} from 'js/utils/env';
 import * as types from 'js/store/mutations-types';
 
+const getCsrfToken = () => {
+	return axios.get(getApiUrl('token'))
```
bognix: I think we are slowly moving toward using async/await. Consider changing it here.

resources/assets/js/plugins/axios.js
bognix: You have the request object available in interceptor. You can check if request was issued to CSRF token endpoint

resources/assets/js/plugins/axios.js
```diff
@@ -29,10 +42,20 @@ export default (Vue, {store, router}) => {
...
 				}
 			}
+
 			if (error.response && error.response.status === 401) {
 				window.location.replace('/login');
 			}
 
+			if (error.config && error.response && error.response.status === 419) {
+				return getCsrfToken().then((token) => {
+					let config = error.config;
```
bognix: const is perfectly fine here.

resources/assets/js/plugins/axios.js
```diff
@@ -29,10 +42,20 @@ export default (Vue, {store, router}) => {
...
+					config.headers['X-CSRF-TOKEN'] = token;
```
bognix: do you need to set it here since you are setting it as default value one line below?

routes/papi.php
```diff
@@ -25,6 +25,9 @@ function api_action($type, $method)
 	$r = config('papi.resources');
 
 	Route::group(['middleware' => ['admin']], function () use ($r) {
+		// CSRF token
+		Route::get("token", 'TokenApiController@getToken');
```
bognix: Why you put it under admin middleware? Will it work for regular users?

resources/assets/js/plugins/axios.js
```diff
@@ -29,10 +42,20 @@ export default (Vue, {store, router}) => {
...
+					window.axios.defaults.headers.common['X-CSRF-TOKEN'] = token;
+					return axios.request(config);
```
kvas-damian: I think we should protect from getting into infinite loop. Maybe retry counter in the header?

resources/assets/js/plugins/axios.js
gitkuba: Nope. It wouldn't work that way, because the request we're repeating is already configured.

resources/assets/js/plugins/axios.js
```diff
@@ -29,8 +39,21 @@ export default (Vue, {store, router}) => {
...
 				}
 			}
+
 			if (error.response && error.response.status === 401) {
 				window.location.replace('/login');
+				$wnl.logger.warning('XHR resulted in 401', error.response);
+			}
+
+			if (error.config.url !== getApiUrl('token') &&
+				error.config && error.response &&
```
bognix: shouldn't error config be checked before `error.config.url`?

routes/papi.php
```diff
@@ -24,7 +24,11 @@ function api_action($type, $method)
 Route::group(['namespace' => 'Api\PrivateApi', 'middleware' => ['api-auth']], function () {
 	$r = config('papi.resources');
 
+	// CSRF token
+	Route::get("token", 'TokenApiController@getToken');
```
bognix: we have a group of requests that don't require active subscription at the bottom of this long list. Could you move it there?

resources/assets/js/plugins/axios.js
bognix: What is more you already use lodash.get in this file, you can make use of it here as well.

resources/assets/js/plugins/axios.js
```diff
@@ -29,8 +39,21 @@ export default (Vue, {store, router}) => {
...
+				error.response.status === 419
+			) {
+				const token = await getCsrfToken();
```
bognix: if this request fails it goes to interceptor once again? Just wanted to confirm :)

resources/assets/js/utils/logger.js
```diff
@@ -32,7 +32,7 @@ export default class Logger {
 		this.level     = envValue('APP_LOG_LEVEL');
 		this.levelCode = Logger.LEVELS[this.level];
 
-		if (this.useExternal(this.levelCode)) {
+		if (!isDev()) {
```
bognix: why this change?

resources/assets/js/plugins/axios.js
```diff
@@ -1,7 +1,17 @@
...
-import { get } from 'lodash';
+import {get} from 'lodash';
+import {getApiUrl} from 'js/utils/env';
 import * as types from 'js/store/mutations-types';
 
+const getCsrfToken = async () => {
+	const response = await axios.get(getApiUrl('token'));
+	const token = get(response, 'data.token');
+	if (!token) {
+		throw new Error('Failed to resolve a valid CSRF token from server.');
```
bognix: if you throw an error here won't it trigger the interceptor? Will it log to sentry automatically?

resources/assets/js/plugins/axios.js
```diff
@@ -29,8 +39,21 @@ export default (Vue, {store, router}) => {
...
+
+			if (error.config.url !== getApiUrl('token') &&
+				error.config && error.response &&
+				error.response.status === 419
+			) {
+				const token = await getCsrfToken();
+				const config = error.config;
+				config.headers['X-CSRF-TOKEN'] = token;
+				window.axios.defaults.headers.common['X-CSRF-TOKEN'] = token;
+				return axios.request(config);
```
kvas-damian: retrying in the interceptor ofter leads to request infinite loop. Besides the problem for a client it can cause some trouble for our server if a few users get into this state. 
What do you think about adding a retry limit?

resources/assets/js/plugins/axios.js
gitkuba: yup

resources/assets/js/utils/logger.js
gitkuba: Previously `Raven` wasn't initialized if `LOG_LEVEL` was set to `debug`, which means that events of all levels were suppressed.

resources/assets/js/plugins/axios.js
gitkuba: Not quite. It had some issues after changing to `async/await`. Thanks for vigilance! :)

resources/assets/js/plugins/axios.js
gitkuba: It won't retry if response status is not `419`. Under some bizarre conditions `/token` endpoint could return 419, therefore we check `error.config.url !== getApiUrl('token')`.

resources/assets/js/plugins/axios.js
gitkuba: I'm not sure how `get()` could help here. Isn't it sufficient to change order of comparison expressions? Please, take a look at recent changes.

PR #1270, opened 2019-02-06
PLAT-959 | Recent terms on content classifier

kvas-damian:


resources/assets/js/components/global/contentClassifier/ContentClassifierEditorRecentTerms.vue
```diff
@@ -0,0 +1,81 @@
...
+			}"
+		>
+			<span class="icon is-small">
+				<i class="fa fa-tag"></i>
+			</span>
+			Dodaj ostatnio użyte pojęcie: <strong>{{lastUsedTerm.tag.name}}</strong>
+		</div>
+		<div
+			v-if="lastUsedTermsSet"
+			@click="canAddLastUsedTermsSet && lastUsedTermsSet.map(term => $emit('attachTaxonomyTerm', term))"
```
kvas-damian: Currently, we call as many requests as we have terms to assign? Do we want to try this approach or create a separate endpoint for attaching multiple terms to multiple items?
On the other hand, we have a batch endpoint to create implemented. Maybe we could reuse it here?

resources/assets/js/components/global/contentClassifier/ContentClassifierEditor.vue
```diff
@@ -232,6 +245,8 @@ export default {
 					ancestors: term.ancestors || this.getAncestorsById(term.id),
 				};
 				this.$emit('taxonomyTermAttached', termToAdd);
+				contentClassifierStore.set(CONTENT_CLASSIFIER_STORE_KEYS.LAST_TERM, termToAdd);
+				contentClassifierStore.set(CONTENT_CLASSIFIER_STORE_KEYS.ALL_TERMS, this.allTaxonomyTerms);
```
bognix: correct me if I'm wrong. This is the list of all taxonomy terms used in the editor?

resources/assets/js/components/global/contentClassifier/ContentClassifierEditorRecentTerms.vue
```diff
@@ -0,0 +1,81 @@
...
+			type: Array,
+			required: true,
+		},
+	},
+	computed: {
+		canAddLastUsedTerm() {
+			return this.lastUsedTerm && !this.items.every(item => item.taxonomyTerms.find(term => term.id === this.lastUsedTerm.id));
+		},
+		canAddLastUsedTermsSet() {
+			return this.lastUsedTermsSet && !this.lastUsedTermsSet.every(lastUsedTerm => this.items.every(item => item.taxonomyTerms.find(term => term.id === lastUsedTerm.id)));
```
bognix: I would abstract this part `this.items.every(item => item.taxonomyTerms.find(term => term.id === lastUsedTerm.id))` to a function. The `lastUserTerm` can be passed as an argument.

resources/assets/js/components/global/contentClassifier/ContentClassifierEditor.vue
kvas-damian: yes, we use all. We didn't agree on anything so I choose a random option. If you feel there is a better option to choose which terms should be presented, let me know.

bognix: the option is fine, the variable name wasn't descriptive enough for me, hence I asked :)

resources/assets/js/components/global/contentClassifier/ContentClassifierEditorRecentTerms.vue
gitkuba: Both options soud good to me.

PR #1271, opened 2019-02-06
PLAT-958 | Add shortcut keys for content item classifier

rogatty:
TODO:

- [x] Add to quiz questions
- [x] Add to slides
- [x] Add tabindex for navigation between taxonomy select and term autocomplete
- [x] Remember the recent taxonomy and use it in the select as a default
- [x] Add shortcuts for recent terms
- [x] Enable shortcuts only if user has access
- [x] Navigate between slides using `[` and `]`
- [x] Navigate between active questions using `[` and `]`
- [x] Fix active instance index when questions page is changed
- [x] When instance is focused, `t` should focus the term autocomplete
- [x] Scroll to editor so it's visible, and not so it's on the top


resources/assets/js/components/course/screens/flashcards/Flashcards.vue
```diff
@@ -55,11 +55,15 @@
 							slot="content"
 							@userEvent="trackUserEvent"
 					/>
-					<wnl-content-item-classifier-editor
-						:key="`cc-editor-${flashcard.id}`"
-						:content-item-id="flashcard.id"
-						:content-item-type="CONTENT_TYPES.FLASHCARD"
-					/>
+					<wnl-activate-with-shortcut-key :key="`cc-editor-${flashcard.id}`">
+						<template slot-scope="activateWithShortcutKey">
```
bognix: do you need this template.

resources/assets/js/components/global/contentClassifier/ContentItemClassifierEditor.vue
```diff
@@ -126,5 +130,10 @@ export default {
 			});
 		},
 	},
+	watch: {
+		'activateWithShortcutKey.isActive'(isActive) {
+			this.expanded = isActive;
```
bognix: this could be handled on the computed level.

resources/assets/js/components/global/contentClassifier/ContentItemClassifierEditor.vue
```diff
@@ -82,6 +82,10 @@ export default {
 		};
 	},
 	props: {
+		activateWithShortcutKey: {
```
bognix: Do you need to pass an object? Isn't isActive enough?

resources/assets/js/components/global/contentClassifier/ContentItemClassifierEditor.vue
bognix: How do you synchronize the store if more than one item is expanded?

resources/assets/js/components/global/ActivateWithShortcutKey.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+	<div>
```
bognix: do you need this div around?

resources/assets/js/components/course/screens/flashcards/Flashcards.vue
rogatty: Yes, I believe so. CCE wasn't rendered without it.

resources/assets/js/components/global/ActivateWithShortcutKey.vue
rogatty: Yes, see https://vuejs.github.io/eslint-plugin-vue/rules/valid-template-root.html

resources/assets/js/components/global/contentClassifier/ContentItemClassifierEditor.vue
rogatty: Updated, now there can only be a single item opened.

resources/assets/js/components/global/contentClassifier/ContentItemClassifierEditor.vue
```diff
@@ -110,18 +113,21 @@ export default {
...
+			return this.getContentItemState({contentItemType: this.contentItemType, contentItemId: this.contentItemId}) === REQUEST_STATES.ERROR;
+		},
 	},
 	methods: {
 		...mapMutations('contentClassifier', {
 			attachTerm: CONTENT_CLASSIFIER_ATTACH_TERM,
 			detachTerm: CONTENT_CLASSIFIER_DETACH_TERM
 		}),
+		updateIsActive(isActive) {
+			this.$parent.$emit('updateIsActive', isActive);
```
bognix: this pattern we should avoid. We don't want tightly couple child with the parent.

rogatty: It's based on https://github.com/vuejs/vue/issues/4332#issuecomment-263444492 and I didn't find any better solution to communicate between a slot child and a slot container.

resources/assets/js/components/global/taxonomies/TaxonomyTermAutocomplete.vue
```diff
@@ -88,10 +100,24 @@ export default {
...
+		},
 		onSelect(item) {
 			this.search = '';
 			this.$emit('change', item);
 		},
 	},
+	watch: {
+		async isFocused(isFocused) {
+			if (isFocused) {
+				this.$refs.input.focus();
```
kvas-damian: Did you try using directives for that? https://vuejs.org/v2/guide/custom-directive.html
I never used it before, but sounds like a possible solution here.

resources/assets/js/store/modules/contentClassifier.js
```diff
@@ -84,6 +85,9 @@ const mutations = {
 			...reducedState
 		});
 	},
+	[mutationsTypes.CONTENT_CLASSIFIER_SET_RECENT_TAXONOMY_ID](state, taxonomyId) {
+		set(state, 'recentTaxonomyId', taxonomyId);
```
kvas-damian: Imo it should be stored in LocalStorage

resources/assets/js/components/global/contentClassifier/ContentItemClassifierEditor.vue
```diff
@@ -18,11 +18,17 @@
 				<wnl-content-classifier-editor
 					v-if="hasContentItem"
 					:items="[contentItem]"
+					:on-attach-last-used-term-triggered="onAttachLastUsedTermTriggered"
```
kvas-damian: looks like event not props.

resources/assets/js/components/course/screens/flashcards/Flashcards.vue
```diff
@@ -46,21 +46,28 @@
...
-				</template>
+						/>
+						<wnl-content-item-classifier-editor
+							:is-active="activateWithShortcutKey.isActive"
+							:is-focused="activateWithShortcutKey.isFocused"
+							:content-item-id="flashcard.id"
+							:content-item-type="CONTENT_TYPES.FLASHCARD"
+							@updateIsActive="activateWithShortcutKey.onUpdateIsActive"
+							@editorCreated="activateWithShortcutKey.onComponentCreated"
+							@editorDestroyed="activateWithShortcutKey.onComponentDestroyed"
```
kvas-damian: I don't like the fact that we need to pass 5 additional props to make shortcuts work. I think we're doing something wrong way. I need to investigate that.

resources/assets/js/components/global/ActivateWithShortcutKey.vue
```diff
@@ -0,0 +1,57 @@
...
+				this.$shortcutKeySetActiveInstance(`activate-with-shortcut-key-${this._uid}`);
+				this.isActive = true;
+			} else {
+				this.$shortcutKeyResetActiveInstance();
+			}
+		},
+		onComponentCreated() {
+			this.$shortcutKeyRegister({
+				uid: `activate-with-shortcut-key-${this._uid}`,
+				onActivate: async () => {
```
kvas-damian: Instead of callbacks, let's use getters and watchers (as usual in vue)

resources/assets/js/components/global/ActivateWithShortcutKey.vue
```diff
@@ -0,0 +1,57 @@
...
+					this.isActive = true;
+
+					await nextTick();
+					scrollToElement(this.$el, 150, 500);
```
kvas-damian: 2 last params are optional

resources/assets/js/components/global/contentClassifier/ContentClassifierEditor.vue
```diff
@@ -54,8 +54,12 @@
 			</ul>
 
 			<wnl-content-classifier-editor-recent-terms
+				:trigger-attach-last-used-term="triggerAttachLastUsedTerm"
+				:trigger-attach-last-used-terms-set="triggerAttachLastUsedTermsSet"
 				:items="items"
 				@attachTaxonomyTerm="onAttachTaxonomyTerm"
+				@attachLastUsedTermTriggered="onAttachLastUsedTermTriggered"
+				@attachLastUsedTermsSetTriggered="onAttachLastUsedTermsSetTriggered"
```
kvas-damian: How can we avoid padding it up and down?

rogatty: 1. Create a store that would control this.
2. Move logic to `ContentClassifierItemEditor`

I don't like either of these.

resources/assets/js/components/global/contentClassifier/ContentItemClassifierEditor.vue
```diff
@@ -134,6 +178,32 @@ export default {
...
+	},
+	mounted() {
+		if (this.canAccess) this.$emit('editorCreated');
+	},
+	beforeDestroy() {
+		this.$emit('editorDestroyed');
+	},
+	watch: {
+		async isFocused() {
+			if (this.isFocused) {
```
kvas-damian: this part is tricky, we can easily get into a state where `isFocused` is set to true and the element isn't focused (when a user clicks outside of the component).

resources/assets/js/components/global/taxonomies/TaxonomyTermAutocomplete.vue
kvas-damian: The PoC was developed in https://github.com/bethinkpl/wnl-platform/pull/1287 but we agree to keep the current solution.

resources/assets/js/components/questions/ActiveQuestion.vue
```diff
@@ -210,6 +218,32 @@ export default {
...
+				this.$shortcutKeySetActiveInstance('active-question');
+				this.isContentItemClassifierEditorActive = true;
+			} else {
+				this.$shortcutKeyResetActiveInstance();
+				this.isContentItemClassifierEditorActive = false;
+			}
+		},
+		onContentItemClassifierEditorCreated() {
+			this.$shortcutKeyRegister({
+				uid: 'active-question',
```
kvas-damian: constant

resources/assets/js/components/questions/QuestionsSolving.vue
```diff
@@ -211,7 +206,7 @@
 			.wnl-quiz-question
 				margin: 0
 
-	.quiz-question__content-item-classifier-editor
+	.questions-list__content-item-classifier-editor
```
kvas-damian: not used?

resources/assets/js/components/quiz/QuizWidget.vue
```diff
@@ -55,12 +50,6 @@
 					@selectAnswer="selectAnswer"
 					@answerDoubleclick="onAnswerDoubleClick"
 				></wnl-quiz-question>
-				<wnl-content-item-classifier-editor
```
kvas-damian: Styles not used now

resources/assets/js/slideshow.js
```diff
@@ -303,6 +303,10 @@ function keyDown(e, parent) {
 		toggleBookmark(parent);
 		break;
 	}
+
+	if (['t', '[', ']'].indexOf(e.key) > -1) {
```
kvas-damian: `includes`

resources/assets/js/store/modules/activateWithShortcutKey.js
```diff
@@ -0,0 +1,134 @@
...
+		if (index > -1) {
+			if (state.activeInstanceIndex > -1) {
+				state.instances[state.activeInstanceIndex].onDeactivate();
+			}
+
+			commit(types.ACTIVATE_WITH_SHORTCUT_KEY_SET_ACTIVE_INSTANCE_INDEX, index);
+			state.instances[index].onActivate(state.instances[index].uid);
+		}
+	},
+	setFirstInstanceAsActive({commit, state}) {
```
kvas-damian: It should be not exposed as a public method as we don't check there is no other active instance.

resources/assets/js/components/global/contentClassifier/ContentItemClassifierEditor.vue
```diff
@@ -110,18 +128,42 @@ export default {
...
+		onKeyDown(event) {
+			if (this.isFocused && !this.$shortcutKeyIsEditable(event.target)) {
+				switch (event.key) {
+				case 't':
+					// Disable global shortcut
+					event.stopImmediatePropagation();
+					this.triggerFocusTaxonomyTermAutocomplete = true;
+					break;
+
+				case 'r':
```
kvas-damian: these shortcuts won't work in admin editor for multiple items.

resources/assets/js/components/global/taxonomies/TaxonomyTermAutocomplete.vue
```diff
@@ -92,5 +98,16 @@ export default {
 			this.$emit('change', item);
 		},
 	},
+	watch: {
+		async isFocused(isFocused) {
+			if (isFocused) {
+				this.$refs.autocomplete.$refs.input.focus();
```
bognix: this chain of refs is a bit crazy.

kvas-damian: An alternative to that is to pass `isFocused` as a props and immediately emit an event to change it's value. I think it's the right moment to agree on one standard.

PR #1273, opened 2019-02-10
Use NFS to improve docker performance on Mac OS

gitkuba:
Solution based on:
https://medium.com/@sean.handley/how-to-set-up-docker-for-mac-with-native-nfs-145151458adc

We'll need to add `NFS_SOURCE_DIR` var.

Those, who think they could benefit from switching to nfs should also change their `LOCAL_CODE_PATH` to `nfsmount` and execute the `setup_native_nfs_docker_osx.sh` script in main code dir, as described in the article.

On my machine I was able to observe following results:
http://platform.local/papi/v2/course_structure_nodes/1?include=groups,lessons,courses
before: 3.7 - 4s
after: ~600ms
(from network tab)

http://platform.local/admin/app
before: 2-3s
after: ~100ms
(server time from debug bar)

There is also change from bind mount to volume in mysql container configuration, which has given me almost 50% time reduction for mysql queries. Unluckily I wasn't able to use nfs for mysql due to permissions issues. This means your DB data will be lost and you'll need to load a dump again, but then it should be preserved even if you remove mysql container (Well, unless you'll do it using `-v` flag, like `docker rm <container> -v` or remove the volume explicitly with `docker volume rm`).


rogatty:
Nice but we should put the instructions somewhere. README, maybe?

PR #1274, opened 2019-02-11
PLAT-958 | Add keyboard support to Autocomplete component

kvas-damian:
* integrate search input and Autocomplete component
* add keyboard support yo Autocomplete component
* fix bug on chat input when selecting from autocomplete using a keyboard

To test:
* [ ] all autocomplete usages
* [ ] moderators autocomplete
* [ ] chat input autocomplete
* [ ] moderators feed


resources/assets/js/components/global/AutocompleteList.vue
```diff
@@ -0,0 +1,80 @@
...
+
+		&__text
+			padding: 5px 10px
+</style>
+
+<script>
+import autocompleteNav from 'js/mixins/autocomplete-nav';
+
+export default {
+	props: ['items', 'onItemChosen', 'isDown'],
```
bognix: could you add props validation?

resources/assets/js/components/global/Autocomplete.vue
```diff
@@ -1,81 +1,74 @@
...
+		WnlAutocompleteList,
+	},
+	methods: {
+		onKeyDown(evt) {
+			if ([KEYS.arrowUp, KEYS.arrowDown, KEYS.esc, KEYS.enter].includes(evt.keyCode)) {
+				evt.preventDefault();
+				evt.stopPropagation();
+			}
+
+			this.$refs.autocomplete.onKeyDown(evt);
```
rogatty: We call a child component method here, right? @kvas-damian @bognix do we want to make it a standard solution? In https://github.com/bethinkpl/wnl-platform/pull/1271 I pass a prop to the child, set a watcher and then emit an event back to the parent to achieve the same result as here. Search for e.g. `triggerBlur`. If we decide this is a better solution, then I'll gladly remove all the boilerplate. But I thought using `$refs` isn't recommended?

bognix: I agree. Actually what could be done here is to pass active item index as prop to annotationsList. This way you won't have to neither call `$refs.onKeyDown` nor passing props and setting watches to clear the state. However, the change I mention requires refactoring of few other components.

kvas-damian: @rogatty I moved the keyboard handling higher and now we don't need to call methods on refs. Thanks for pointing that.

resources/assets/js/components/moderators/ModeratorsAutocomplete.vue
```diff
@@ -55,103 +33,48 @@
...
-		show: {
-			type: Boolean,
-			default: false
-		},
-		initialValue: {
-			type: String,
-			default: ''
+		selected: {
+			type: Object,
+			default: () => {},
```
bognix: `default: () => ({})`

PR #1278, opened 2019-02-12
PLAT-982 | Content Classifier select items

kvas-damian:


resources/assets/js/admin/components/contentClassifier/ContentClassifier.vue
```diff
@@ -22,8 +25,16 @@
 								v-for="item in groupedFilteredContent[contentType]"
 								:key="item.id"
 								class="content-classifier__result-item"
+								:class="{'is-active': selectedItemIds.includes(item.id)}"
+								@click="toggleSelected(item)"
 							>
 								<component :is="meta.component" :item="item"/>
+								<span
+									v-if="selectedItemIds.includes(item.id)"
```
bognix: I wonder if instead of `v-if` you could just hide this element using proper selector and `display:none` so you won't duplicate the check in two places.

PR #1283, opened 2019-02-13
[PLAT-979] Improve API tests performance

gitkuba:
- Most time consuming action is saving notifications when we have lots of users in DB. Takes about 80% of execution time for whole test set.
- `MAIL_DRIVER=log` = 2s saved on each mail sent. 
- Hardcode hash instead of using `bcrypt()` - speed gain is fraction of a second for single usage, but it's 18216x faster. Taken from this article: https://driesvints.com/blog/two-tips-to-speedup-your-laravel-tests/
- `SCOUT_DRIVER=null` to omit indexing and save another 30s.
- call `flushEventListeners()` on various models, depending on context to avoid side effects of creating new model instances.

Testing hotspots:
- [x] check `QnaQuestionPosted` event is triggered (notifications are sent etc.)


tests/Api/Course/SlidesTest.php
```diff
@@ -29,7 +32,7 @@ public function post_slide()
 		$response = $this
 			->actingAs($user)
 			->json('POST', $this->url('/slides'), $data);
-
+		$response->dump();
```
kvas-damian: what's that?

PR #1284, opened 2019-02-13
PLAT-946 | Hide taxonomy edit form by default

rogatty:
https://bethink.atlassian.net/browse/PLAT-946


resources/assets/js/admin/components/taxonomies/TaxonomyEditor.vue
```diff
@@ -70,13 +77,22 @@ export default {
...
+	},
 	computed: {
 		isEdit() {
 			return this.id !== 'new';
 		},
 		method() {
 			return this.isEdit ? 'put' : 'post';
 		},
+		formData() {
+			return this.$store.getters['form/getData'](this.formName);
```
kvas-damian: Maybe, instead of hooking into internals of Form component  we could emit an event which contains formData? Similar to `submitSuccess` event.

kvas-damian: Also this way you update the header when editing an input. It's confusing, because you didn't submit this change.

rogatty: @kvas-damian updated

PR #1291, opened 2019-02-14
PLAT-919 | Stop logging nonactionable socket errors

rogatty:
https://bethink.atlassian.net/browse/PLAT-919


resources/assets/js/plugins/chat-connection.js
```diff
@@ -33,11 +33,21 @@ const createEventsQueue = () => {
 const WnlSocket = {
 	install(Vue, {store}) {
 		const onSocketError = (error) => {
+			// Happens e.g. when server is restared or session expires
+			// Socket.io handles it by reconnecting or creating new session
```
kvas-damian: Do we log errors on new session creation?

rogatty: I'm not sure how to check it. Do you have any specific error in mind?

kvas-damian: I was referring to `onSocketConnectionError` which isn't logging anything.

PR #1292, opened 2019-02-15
[PLAT-969] Introduce PHP static analysis tool

gitkuba:


app/Console/Commands/MarkPaymentsAsFailed.php
```diff
@@ -44,11 +44,13 @@ public function handle()
 		if (!empty($paymentId)) {
 			$payment = Payment::find($paymentId);
 			if (empty($payment)) {
-				return $this->warn('Invalid payment id');
+				$this->warn('Invalid payment id');
+				die;
```
rogatty: Why `die`? Can't we just `return`?

app/Http/Controllers/Api/PrivateApi/ReactionsApiController.php
```diff
@@ -64,7 +64,7 @@ public function postMany(Request $request)
 					'user' => $user->id]
 				]);
 
-				return $this->respondInternalError($e->getMessage());
+				return $this->respondInternalError();
```
rogatty: We could also use:
```
return $this
			->setMessage($e->getMessage())
			->setStatusCode(500)
			->json();
```
if the message is valuable for us.

app/Console/Commands/MarkPaymentsAsFailed.php
gitkuba: Function `warn()` doesn't return anything, which pisses of the linter. 🤷‍♂️

app/Providers/DiscussableServiceProvider.php
```diff
@@ -8,13 +8,17 @@
 class DiscussableServiceProvider extends ServiceProvider {
 	public function register() {
 		Blueprint::macro('discussable', function() {
-			$this->tinyInteger('is_discussable')->default(0);
-			$this->integer('discussion_id')->nullable();
+			/** @var mixed */
+			$table = $this;
```
rogatty: What's wrong with the original code?

rogatty:
Did you try to reverse https://github.com/bethinkpl/wnl-platform/pull/1263/commits/adddc59ca92d4da5d5da42b1b45e1c6da5602cd7 and see if the linter fails?

app/Console/Commands/MarkPaymentsAsFailed.php
rogatty: Yeah, but you can use:
```
$this->warn('Invalid payment id');
return;
```

gitkuba:
> Did you try to reverse [adddc59](https://github.com/bethinkpl/wnl-platform/commit/adddc59ca92d4da5d5da42b1b45e1c6da5602cd7) and see if the linter fails?

Yes, it has been caught.

app/Providers/DiscussableServiceProvider.php
gitkuba: Well, problem is, class we're in doesn't have funcitons we call. So it's kind of a hack to avoid linter error 🤷‍♂️ BTW it's the least dirty, we came up with.

rogatty: Ah, seems like this is the one: https://github.com/nunomaduro/larastan/issues/42

rogatty: Worth putting the link in a comment, IMO.

app/Http/Controllers/Api/PrivateApi/ReactionsApiController.php
gitkuba: We have it in sentry. I see no point of sending it to a client. Besides that, if i correctly interpret this code, it's a kind of debugger(?) To tell how ofter a duplicate entry occurs (?) Maybe we should ask @bognix.

bognix: No, it's not a debugger.

app/Console/Commands/MarkPaymentsAsFailed.php
gitkuba: oh, right 😆

app/Events/Chat/PrivateMessageSent.php
```diff
@@ -12,14 +12,4 @@ class PrivateMessageSent extends Event
 	use Dispatchable,
 		InteractsWithSockets,
 		SerializesModels;
-
-	/**
-	 * Create a new event instance.
-	 *
-	 * @return void
-	 */
-	public function __construct()
```
kvas-damian: Are you sure you want to call parent constructor now?

app/Models/Keyword.php
```diff
@@ -10,6 +10,6 @@ class Keyword extends Model
 
 	public function annotation()
 	{
-		return $this->belognsTo('App\Models\Annotation');
+		return $this->belongsTo('App\Models\Annotation');
```
kvas-damian: Nice catch!

larastan.neon
```diff
@@ -0,0 +1,8 @@
+includes:
+    - ./vendor/nunomaduro/larastan/extension.neon
+parameters:
+    level: 0
```
kvas-damian: Do we want to use `level: 0` or 1 as defined in `travis.yml`?

app/Events/Chat/PrivateMessageSent.php
rogatty: I think this was added by an accident, PHPStorm also highlights that the parent constructor wasn't called.

larastan.neon
rogatty: Ah, this one is actually ignored because `php artisan code:analyse` always overrides it. I'll remove it.

app/Http/Controllers/Admin/UploadSlidesController.php
```diff
@@ -13,7 +13,7 @@ public function index(){
 	}
 
 	public function handle(Parser $parser, Request $request) {
-		$parser->parse($request->get('slides'), null, (bool) $request->get('enable_slides_matching'));
+		$parser->parse($request->get('slides'), null, null, (bool) $request->get('enable_slides_matching'));
```
rogatty: Bug caught 🐛

app/Http/Controllers/Api/PrivateApi/TagsApiController.php
```diff
@@ -70,15 +70,9 @@ public function delete($id) {
 			]);
 		}
 
-		if ($tag->hasProtectedTaggable()) {
```
rogatty: Bug caught 🐛

app/Listeners/ContentUpdatesGate.php
```diff
@@ -2,20 +2,24 @@
...
 	 *
-	 * @param  $event
+	 * @param Event $event
 	 *
 	 * @return void
 	 */
 	public function handle($event)
 	{
-		$event->transform();
+		if (method_exists($event, 'transform')) {
```
rogatty: We do the same in https://github.com/bethinkpl/wnl-platform/blob/ceac2a3c588be8c338bba84e35efa0a3e209501b/app/Listeners/UserNotificationsGate.php#L114-L116

app/Models/Concerns/WithReactions.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace App\Models\Concerns;
```
rogatty: I'm not sure if this is the correct namespace for interfaces 🤔

app/Models/Slide.php
```diff
@@ -71,7 +73,8 @@ public function setContentAttribute($value)
 
 	public function setSnippetAttribute($value)
 	{
-		$this->attributes['snippet'] = json_encode(Parser::createSnippet($value));
+		$snippet = (new Parser())->createSnippet($value);
```
rogatty: `createSnippet` is an instance method.

app/Notifications/Media/LiveChannel.php
```diff
@@ -19,7 +19,7 @@ public function send($notifiable, $notification)
 		$message = new BroadcastMessage($notification->event->data);
 
 		$event = new LiveNotificationCreated(
-			$notifiable, $notification, is_array($message) ? $message : $message->data
+			$notifiable, $notification, $message->data
```
rogatty: It's always an instance of `BroadcastMessage`

app/Models/Concerns/WithReactions.php
gitkuba: if i'm not mistaken, in laravel terminology `Concerns` is used for traits and `Contracts` for interfaces 🤷‍♂️

app/Console/Commands/TaxonomizeTags.php
```diff
@@ -45,7 +45,7 @@ public function handle()
 		$tags = $this->argument('tags');
 
 		$taxonomy = Taxonomy::find($taxonomyId);
-		$parentTag = $parentTagId !== 0 ? Tag::find($parentTagId) : 0;
+		$parentTag = $parentTagId !== '0' ? Tag::find($parentTagId) : null;
```
kvas-damian: I think this condition is broken. What about handling an empty string?

app/Events/Chat/PrivateMessageSent.php
kvas-damian: I also hope it was by accident, but it's possible that constructor will cause some unexpected effects on the object...

app/Console/Commands/TaxonomizeTags.php
gitkuba: OK, cool, `$parentTagId` is indeed a string. Makes sense.

kvas-damian: But we can skip finding when we pass an empty string.

app/Events/Chat/PrivateMessageSent.php
gitkuba: I don't think we're using this class anyway 🤷‍♂️

PR #1293, opened 2019-02-15
PLAT-950 | Unify editor for nested set

bognix:
- clear node field when user adds node to root level


resources/assets/js/admin/components/structure/StructureNodeEditorEdit.vue
```diff
@@ -36,41 +63,36 @@ export default {
...
+		onSelectStructurable(structurable) {
+			this.structurable = structurable;
+		},
+	},
+	created() {
+		this.parent = this.getAncestorsById(this.node.id).slice(-1)[0];
+		this.structurable = this.node.structurable;
 	},
+	watch: {
+		term() {
```
bognix: update watch

resources/assets/js/admin/components/taxonomies/TaxonomyTermEditorAdd.vue
```diff
@@ -1,93 +1,103 @@
 <template>
 	<div>
 		<wnl-taxonomy-term-editor-current-term
```
bognix: I'm tempted to create this component for both courseStructure and taxonomyTerms however it's not trivial because they use different fields in nodes to get the display name.

I don't want to extend the scope of this ticket even more so maybe let's leave it as it is, but I'll keep an eye on this part :)

resources/assets/js/admin/components/taxonomies/TaxonomyTermEditorEdit.vue
```diff
@@ -3,12 +3,34 @@
 		<wnl-taxonomy-term-editor-current-term :term="term">
 			Edytujesz pojęcie:
 		</wnl-taxonomy-term-editor-current-term>
-		<wnl-taxonomy-term-editor-form
-			submit-label="Zapisz"
+		<wnl-nested-set-editor-form
```
bognix: There are some pieces shared between TaxonomyTermEditorEdit and TaxonomyTermEditorAdd.
I'm thinking about creating a shared component for those features however I'm worried another level of components may decrease readability.

I'll leave it as it is but I'm open for discussion.

resources/assets/js/admin/components/nestedSet/NestedSetEditor.vue
```diff
@@ -0,0 +1,57 @@
...
+	<div class="nested-set-editor">
+		<div class="nested-set-editor__panel is-left">
+			<div class="nested-set-editor__panel__header">
+				<slot name="header"></slot>
+			</div>
+			<slot name="nodesList" v-if="!isLoading"/>
+			<wnl-text-loader v-else />
+		</div>
+		<div class="nested-set-editor__panel is-right">
+			<slot name="panelRight"></slot>
```
rogatty: From what I see the convention is to use dashes in slot names.

resources/assets/js/admin/components/structure/StructureNodeEditorAdd.vue
```diff
@@ -1,83 +1,82 @@
...
+		</wnl-nested-set-editor-form>
+	</div>
 </template>
 
 <script>
 import {mapActions, mapGetters, mapState} from 'vuex';
 
-import WnlStructureNodeEditorForm from 'js/admin/components/structure/StructureNodeEditorForm';
-import scrollToNodeMixin from 'js/admin/mixins/scroll-to-node';
+import WnlNodeAutocomplete from 'js/admin/components/structure/StructureNodeEditorNodeAutocomplete';
```
rogatty: I believe that until now we were keeping var names consistent with the imported component name (plus prefix). Don't we want to use `WnlStructureNodeEditorNodeAutocomplete` here?

resources/assets/js/admin/components/structure/StructureNodeEditorEdit.vue
```diff
@@ -1,11 +1,26 @@
 <template>
-	<wnl-structure-node-editor-form
+	<wnl-nested-set-editor-form
 		v-if="node"
-		submit-label="Zapisz"
+		parent-title="Gałąź nadrzędna"
+		parent-subtitle="Pozostaw puste, aby dodać pojęcie na 1. poziomie taksonomii"
```
rogatty: It's not a taxonomy editor.

resources/assets/js/admin/components/structure/StructureNodeEditorAdd.vue
```diff
@@ -1,83 +1,82 @@
...
-		submit-label="Dodaj"
-		:on-save="onSave"
-		:courseId="courseId"
-		:node="node"
-		@parentChange="onParentChange"
-	/>
+	<div>
+		<wnl-nested-set-editor-form
+			parent-title="Gałąź nadrzędna"
+			parent-subtitle="Pozostaw puste, aby dodać pojęcie na 1. poziomie taksonomii"
```
rogatty: It's not a taxonomy editor.

resources/assets/js/admin/components/structure/StructureNodeEditorEdit.vue
```diff
@@ -36,41 +63,36 @@ export default {
...
+				course_id: this.courseId,
+			};
+			return this.updateNode(node);
+		},
+		onSelectStructurable(structurable) {
+			this.structurable = structurable;
 		},
 	},
+	created() {
+		this.parent = this.getAncestorsById(this.node.id).slice(-1)[0];
```
rogatty: `this.node` can be `null` so this will fail.

resources/assets/js/admin/components/taxonomies/TaxonomyTermEditorEdit.vue
```diff
@@ -21,67 +43,79 @@
...
+				parent_id: this.parent ? this.parent.id : null,
+			};
+			return this.updateTerm(term);
+		},
+		onSelectTag(tag) {
+			this.tag = tag;
 		},
 	},
+	created() {
+		this.description = this.term.description;
```
rogatty: `this.term` can be `null` so this will fail.

rogatty:
@bognix can you point to the actual bug fix?

resources/assets/js/admin/components/nestedSet/NestedSetEditorForm.vue
```diff
@@ -0,0 +1,128 @@
...
+		},
+		submitLabel: {
+			type: String,
+			default: 'Zapisz'
+		},
+		submitDisabled: {
+			type: Boolean,
+			default: false
+		},
+		vuexModule: {
```
kvas-damian: `vuexModuleName`

resources/assets/js/admin/components/nestedSet/NestedSetEditorForm.vue
```diff
@@ -0,0 +1,128 @@
...
+			required: true
+		}
+	},
+	methods: {
+		...mapActions(['addAutoDismissableAlert']),
+		action(action, payload = {}) {
+			return this.$store.dispatch(`${this.vuexModule}/${action}`, payload);
+		},
+		validateParent(parent, source) {
+			if (parent && this.action('getAncestorsById', parent.id).find(t => t.id === source.id)) {
```
kvas-damian: It isn't an action

resources/assets/js/admin/components/taxonomies/TaxonomyTermEditorEdit.vue
kvas-damian: I was also thinking about having this as a shared component between those two. on the other hand, I'm not sure if add and edit will always behave the same...

kvas-damian:
Overal, it looks better than previously. I'm still thinking about form add and edit sharing, but have no strong opinion about it.

resources/assets/js/admin/components/structure/StructureNodeEditorAdd.vue
bognix: We can - I copied and pasted.

resources/assets/js/admin/components/taxonomies/TaxonomyTermEditorEdit.vue
bognix: Let's leave it as it is right now then. I decided to unified this way because we share the logic between taxonomy editor and course structure editor. With next tickets we can make further decisions.

resources/assets/js/admin/components/nestedSet/NestedSetEditorForm.vue
```diff
@@ -0,0 +1,135 @@
...
+		}
+	},
+	methods: {
+		...mapActions(['addAutoDismissableAlert']),
+		getter(getter, payload) {
+			const storeGetter = this.$store.getters[`${this.vuexModuleName}/${getter}`];
+			if (typeof storeGetter === 'function') {
+				return storeGetter(payload);
+			}
+			return storeGetter(payload);
```
rogatty: Seems redundant.

bognix: Yes but I didn't want to omit that now and in few months from now someone will just assume that non-function getter works and boom! error!

rogatty: Hmm, but you call it as a function anyway, right? It's always `storeGetter(payload)`.

bognix: oh, that's my bad :D thank you for pointing this out!

PR #1297, opened 2019-02-19
[PLAT-1008] Update user stats script; Add new fields

gitkuba:


app/Console/Commands/ExportUserStatistics.php
```diff
@@ -40,54 +42,86 @@ public function __construct()
...
+
 		return;
 	}
 
 	protected function getUserGroups($minDate, $maxDate, $productIds)
 	{
-		$firstGroup = collect();
-		$secondGroup = collect();
-		$thirdGroup = collect();
+		$oldProducts = [1, 2, 5, 6, 9, 10];
```
kvas-damian: we could make a query to get this list instead of hardcoding, or we could pass it as a parameter.

app/Console/Commands/ExportUserStatistics.php
```diff
@@ -96,33 +130,70 @@ protected function getUserGroups($minDate, $maxDate, $productIds)
...
+				$userRecord['userSectionsProgressPercentage'] >= 20 ||
+				$userRecord['userQuizQuestionsSolved'] >= 300 ||
+				$userRecord['time'] >= 100;
+
+			$G5 = !$G1 && !$G2 && !$G3 && !$G4;
+
+			$userClassification = [
+				'N-G1' => $newUser && $G1,
+				'N-G2' => $newUser && ($G1 || $G2),
+				'N-G3' => $newUser && ($G1 || $G3),
```
kvas-damian: I don't think we need to check `$G1` it's already included in `$G3`

PR #1298, opened 2019-02-19
PLAT-942 | Order by order_number by default

rogatty:
https://bethink.atlassian.net/browse/PLAT-942

```
mysql> SELECT DISTINCT TABLE_NAME
    ->     FROM INFORMATION_SCHEMA.COLUMNS
    ->     WHERE COLUMN_NAME IN ('order_number')
    ->         AND TABLE_SCHEMA='wnl';
+--------------------------+
| TABLE_NAME               |
+--------------------------+
| flashcards_set_flashcard |
| groups                   |
| lessons                  |
| order_instalments        |
| presentables             |
| product_instalments      |
| screens                  |
| sections                 |
| subsections              |
| tags_taxonomy            |
+--------------------------+
10 rows in set (0.01 sec)
```
I skipped `order_instalments` and `product_instalments` because 1. I didn't want to touch checkout and 2. I'm not sure if `order_number` doesn't mean something else in their contexts.


app/Jobs/CalculateCoursePlan.php
```diff
@@ -153,8 +153,7 @@ function ($date) {
 		}
 
 		$builder = $this->user->lessonsAvailability()
-			->orderBy('group_id')
-			->orderBy('order_number');
+			->orderBy('group_id');
```
kvas-damian: Please double check that we preserve the same order now. `order by group_id, order_number` is not the same as `order by order_number, group_id`

rogatty: Fixed in https://github.com/bethinkpl/wnl-platform/pull/1298/commits/ff6cb67d226c9881e13585e65d56dfd057b33a8f

PR #1302, opened 2019-02-20
laravel logs into stdout

r0bj:
1. Change newrelic link to use archive path instead of release - it wouldn't require to change the link on every newrelic release
2. Use /dev/stdout as a laravel log, storing logs in files inside containers is not a good idea:
- containers use ephemeral storage - logs disappear on container restart
- no logs file rotation - abusive app can fill up all storage without clear indication what happened
- every app in container should send logs to stdout https://12factor.net/
- logs from containers (stdout) are gathered in k8s


Dockerfile.php
```diff
@@ -33,14 +33,14 @@
...
+RUN curl -L https://download.newrelic.com/php_agent/archive/8.5.0.235/newrelic-php5-8.5.0.235-linux-musl.tar.gz | tar -C /tmp -zx && \
 NR_INSTALL_USE_CP_NOT_LN=1 NR_INSTALL_SILENT=1 /tmp/newrelic-php5-*/newrelic-install install && \
 rm -rf /tmp/newrelic-php5-* /tmp/nrinstall*
 
 WORKDIR /www/current
 
 COPY --from=js-build /src/. .
-RUN touch storage/logs/laravel.log
 RUN chown -R 82:82 /www /run /tmp
+RUN ln -s /dev/stdout storage/logs/laravel.log
```
rogatty: Let's skip this layer and write logs to stdout directly.

r0bj: @rogatty can you configure laravel that way?

rogatty: Yup, on it.

rogatty:
I'm not sure if we should merge it yet as the logs are in the following format:
```
[20-Feb-2019 13:10:54] WARNING: [pool www] child 10 said into stdout: "[2019-02-20 13:10:54] dev.ERROR: what!  "
```
and can be trimmed. See https://github.com/docker-library/php/issues/207 for details. There are some workaround but the best solution would be to upgrade PHP to 7.3 which we already have a ticket for.

r0bj:
Sure, we can wait for PHP 7.3.

rogatty:
With PHP 7.3.2 it's:
```
[2019-02-21 15:10:37] dev.ERROR: what!
```

PR #1303, opened 2019-02-20
PLAT-1005 | Default taxonomy in user settings

kvas-damian:


resources/assets/js/components/global/form/Form.vue
```diff
@@ -7,7 +7,7 @@
 			:timestamp="timestamp"
 			@delete="onDelete"
 		></wnl-alert>
-		<slot></slot>
+		<slot :onSubmit="onSubmitForm"></slot>
```
bognix: `:on-submit`

resources/assets/js/components/global/form/Select.vue
```diff
@@ -0,0 +1,59 @@
...
+				class="help is-danger"
+				v-for="(error, index) in getErrors"
+				v-text="error"
+				:key="index"
+			></span>
+		</template>
+	</div>
+</template>
+
+<style lang="sass" rel="stylesheet/sass" scoped>
```
bognix: remove empty styles

resources/assets/js/components/global/form/Select.vue
```diff
@@ -0,0 +1,59 @@
...
+
+</style>
+
+<script>
+import { formInput } from 'js/mixins/form-input';
+
+export default {
+	name: 'TextInput',
```
bognix: fix the name.

resources/assets/js/components/global/form/Select.vue
```diff
@@ -0,0 +1,59 @@
...
+	props: {
```
bognix: those props are almost the same in each component. Since we decided to use mixin maybe we can move them to the mixin?

resources/assets/js/components/user/MySettings.vue
```diff
@@ -15,31 +15,68 @@
...
+
 	</div>
 </template>
 
 <script>
-import { mapGetters, mapActions } from 'vuex';
+import { mapGetters, mapActions, mapState } from 'vuex';
 
-import { Form, Check } from 'js/components/global/form';
+import { Form, Check, Select as WnlFormSelect } from 'js/components/global/form';
```
bognix: how about unifying all components since you are here?

resources/assets/js/consts/settings.js
```diff
@@ -0,0 +1,7 @@
+export const SETTING_NAMES = {
```
rogatty: `USER_ SETTING_NAMES`?

resources/assets/js/components/global/taxonomies/TaxonomyTermSelector.vue
```diff
@@ -92,10 +92,11 @@ export default {
 		try {
 			await this.fetchTaxonomies();
 
-			const lastTaxonomyId = contentClassifierStore.get(CONTENT_CLASSIFIER_STORE_KEYS.LAST_TAXONOMY_ID);
+			await this.setupCurrentUser();
+			const defaultTaxonomyId = this.getSetting(SETTING_NAMES.DEFAULT_TAXONOMY_ID);
 
-			if (lastTaxonomyId) {
-				this.taxonomyId = lastTaxonomyId;
+			if (defaultTaxonomyId) {
```
rogatty: I think we should validate it against the list of taxonomies. You can have an already deleted taxonomy as your default.

resources/assets/js/components/global/form/Form.vue
rogatty: Say no more: https://github.com/bethinkpl/wnl-platform/pull/1306

resources/assets/js/components/global/form/Select.vue
kvas-damian: From my experience props are very different across these components, see `Quill.vue`. I'm don't find a good reason to put it there until we unify the API: `disabled, placeholder, etc...`

PR #1306, opened 2019-02-20
Enable vue/attribute-hyphenation eslint rule

rogatty:
Fortunately, `yarn run lint:fix` did its job 😄


resources/assets/js/admin/components/tags/TaggablesMover.vue
```diff
@@ -1,16 +1,16 @@
 <template>
 	<wnl-form
-		:beforeSubmit="beforeSubmit"
-		:resourceRoute="resourceRoute"
+		:before-submit="beforeSubmit"
+		:resource-route="resourceRoute"
 		@submitSuccess="onSubmitSuccess"
-		hideDefaultSubmit="true"
+		hide-default-submit="true"
```
kvas-damian: `="true"` is unnecessary

kvas-damian: Also, it's passed as a string. The same comment applies to multiple places

PR #1311, opened 2019-02-21
PLAT-1026 | Sync Coupons between platforms

bognix:
### Setup
Firstly, setup your env to run two platforms in parallel: https://github.com/bethinkpl/wnl-platform/wiki/Run-two-platforms-in-parallel

Secondly, add following variables to `.env`:
```
APP_COUPONS_SYNC_URL=
APP_COUPONS_SYNC_TOKEN=
APP_COUPONS_SYNC_IS_SOURCE=
```

- `APP_COUPONS_SYNC_URL` - the url of the second platform that should receive a sync request.
- `APP_COUPONS_SYNC_TOKEN` - the token required to authorize the request
- `APP_COUPONS_SYNC_IS_SOURCE` - boolean value. Whether this platform should notify the other platform about created coupon. Should be true to LEK

In order to make it work on dev we need to define aliases for nginx container:
Firstly, you need to disconnect nginx from the network
```
docker network disconnect wnl-platform_default nginx
```
Secondly, you need to add aliases:
```
docker network connect --alias platform.local --alias platform-copy.local wnl-platform_default nginx
```

### TODO
- [x] The coupon update should happen on queue
- [x] Possible code cleanup
- [x] Unit Tests
- [x] Consider moving the coupon code away from the request url. In case of group discounts the coupon code is the email and there may be some edge cases when it won't work well.
- [x] Use Guzzle/Http library instead of Requests
- [x] Sync coupons on delete


app/Http/Middleware/VerifyCsrfToken.php
```diff
@@ -12,6 +12,6 @@ class VerifyCsrfToken extends BaseVerifier
      * @var array
      */
     protected $except = [
-        //
+        'api/v1/coupons/*'
```
bognix: check if this line is needed. I added it at the very beginning but the source of the problem might be something else.

config/papi.php
```diff
@@ -57,5 +57,6 @@
 		'user-flashcard-notes'    => 'user_flashcard_notes',
 		'roles'                   => 'roles',
 		'course-structure-nodes'  => 'course_structure_nodes',
+		'coupons'                 => 'coupons'
```
bognix: not sure if this should be in `papi.php` since the resource is used in `api.php`

app/Http/Controllers/Api/PublicApi/CouponsApiController.php
```diff
@@ -0,0 +1,44 @@
...
+		return $this->transformAndRespond($coupon);
+	}
+
+	public function put(Request $request)
+	{
+		if ($request->header('BETHINK_COUPON_SYNC_TOKEN') !== env('APP_COUPONS_SYNC_TOKEN')) {
+			return $this->respondUnauthorized();
+		}
+
+		$coupon = Coupon::where('code', $request->route('code'))->first();
```
rogatty: Is it possible that we'll want to update coupons that are already created? If so, this would have to be `firstOrCreate`.

composer.json
```diff
@@ -21,7 +21,8 @@
 		"graham-campbell/throttle": "^7.0",
 		"nao-pon/flysystem-google-drive": "~1.1",
 		"bschmitt/laravel-amqp": "^1.2",
-		"kalnoy/nestedset": "^4.3"
+		"kalnoy/nestedset": "^4.3",
+		"rmccue/requests": "^1.7"
```
rogatty: We already have GuzzleHttp that you can use, see https://github.com/bethinkpl/wnl-platform/blob/cccfbb7327db4dd8aa7521816ad73f4188151b3f/app/Console/Kernel.php#L118

app/Observers/CouponObserver.php
```diff
@@ -0,0 +1,53 @@
...
+		if (empty($coupon->studyBuddy)) {
+			$couponToCreate = $coupon->toArray();
+			unset($couponToCreate['id']);
+
+			$headers = [
+				'Accept' => 'application/json',
+				'Host' => env('APP_COUPONS_SYNC_HOST'),
+				'X-BETHINK-COUPON-SYNC-TOKEN' => env('APP_COUPONS_SYNC_TOKEN'),
+			];
+			Requests::post(env('APP_COUPONS_SYNC_URL') . '/api/v1/coupons', $headers, [
```
rogatty: Will the coupon be removed if this request fails? Because it should.

app/Observers/CouponObserver.php
```diff
@@ -0,0 +1,53 @@
...
+		if (empty($coupon->studyBuddy)) {
+			$couponToUpdate = $coupon->toArray();
+			unset($couponToUpdate['id']);
+
+			$headers = [
+				'Accept' => 'application/json',
+				'Host' => env('APP_COUPONS_SYNC_HOST'),
+				'X-BETHINK-COUPON-SYNC-TOKEN' => env('APP_COUPONS_SYNC_TOKEN'),
+			];
+			Requests::put(env('APP_COUPONS_SYNC_URL') . "/api/v1/coupons/{$coupon->code}", $headers, [
```
rogatty: This should be async.

bognix: it's mentioned in the TODO.

app/Http/Controllers/Api/PublicApi/CouponsApiController.php
bognix: Wasn't sure. Would have to check with our product :)
I think safest option is to not create implicitly here.

composer.json
bognix: I added the TODO

app/Http/Controllers/Api/PublicApi/CouponsApiController.php
bognix: I double checked with Adam and we don't want to create coupons implicitly. Instead we want to log if such situation happens.

app/Events/Coupons/CouponEvent.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace App\Events\Coupons;
+
+abstract class CouponEvent {
```
rogatty: https://stackoverflow.com/a/1814899
>Use an interface when you want to force developers working in your system (yourself included) to implement a set number of methods on the classes they'll be building.
>
>Use an abstract class when you want to force developers working in your system (yourself included) to implement a set numbers of methods and you want to provide some base methods that will help them develop their child classes.

I think an interface (a Laravel contract) would be more appropriate here.

app/Models/Coupon.php
```diff
@@ -2,17 +2,28 @@
...
 
+use App\Events\Coupons\CouponCreated;
+use App\Events\Coupons\CouponDeleted;
+use App\Events\Coupons\CouponUpdated;
 use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Model;
 
 class Coupon extends Model
 {
+	const SYNC_TOKEN_HEADER = "X-BETHINK-COUPON-SYNC-TOKEN";
```
rogatty: It's not related to the model so I don't think it belongs here. I don't have a better idea where to put it, though 🤔

rogatty: Config, maybe?

app/Http/Controllers/Api/PublicApi/CouponsApiController.php
```diff
@@ -0,0 +1,63 @@
...
+
+class CouponsApiController extends ApiController
+{
+	public function __construct(Request $request)
+	{
+		parent::__construct($request);
+		$this->resourceName = config('api.resources.coupons');
+	}
+
+	public function post(Request $request)
```
rogatty: Even if there is a special header, I think we should still validate the input data.

app/Http/Controllers/Api/PublicApi/CouponsApiController.php
```diff
@@ -0,0 +1,63 @@
...
+	{
+		if ($request->header(Coupon::SYNC_TOKEN_HEADER) !== env('APP_COUPONS_SYNC_TOKEN')) {
+			return $this->respondUnauthorized();
+		}
+
+		$coupon = new Coupon($request->coupon);
+		$coupon->removeObservableEvents(['created']);
```
rogatty: Let's add a comment explaining why we're disabling this event.

app/Http/Controllers/Api/PublicApi/CouponsApiController.php
```diff
@@ -0,0 +1,63 @@
...
+		$coupon->save();
+		return $this->transformAndRespond($coupon);
+	}
+
+	public function put(Request $request)
+	{
+		if ($request->header(Coupon::SYNC_TOKEN_HEADER) !== env('APP_COUPONS_SYNC_TOKEN')) {
```
rogatty: I'd move it to a request validator's `authorize()` method.

app/Events/Coupons/CouponUpdated.php
```diff
@@ -0,0 +1,29 @@
...
+		$this->coupon = $coupon;
+	}
+
+	public function shouldSync() {
+		return empty($this->coupon->studyBuddy);
+	}
+
+	public function sync() {
+		$couponToUpdate = $this->coupon->toArray();
+		unset($couponToUpdate['id']);
```
rogatty: We're only using `code` and `times_usable` so how about using a whitelist here instead?

app/Http/Controllers/Api/PublicApi/CouponsApiController.php
```diff
@@ -0,0 +1,63 @@
...
+			return $this->respondUnauthorized();
+		}
+		$updatedCoupon = $request->coupon;
+
+		$coupon = Coupon::where('code', $updatedCoupon['code'])->first();
```
rogatty: Remember to log when there is more than one coupon with this code.

app/Http/Controllers/Api/PublicApi/CouponsApiController.php
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	public function delete($id)
+	{
+		if ($this->request->header(Coupon::SYNC_TOKEN_HEADER) !== env('APP_COUPONS_SYNC_TOKEN')) {
+			return $this->respondUnauthorized();
+		}
+
+		$deletedCoupon = $this->request->coupon;
+		$coupon = Coupon::where('code', $deletedCoupon['code'])->first();
```
rogatty: What if there are two coupons with this code and we remove the wrong one? 🙊

app/Events/Coupons/CouponEvent.php
bognix: I think so two but the shape of this class changed a lot in last two hours.

app/Http/Controllers/Payment/PersonalDataController.php
```diff
@@ -136,13 +136,14 @@ protected function generateStudyBuddy($order)
...
 
-		$coupon->products()->attach(
-			Product::whereIn('slug', ['wnl-online', 'wnl-online-onsite'])->get()
-		);
-
 		$order->studyBuddy()->create([
 			'code' => $coupon->code,
 		]);
+
+		$coupon->save();
```
rogatty: I don't see `$coupon->studyBuddy` being attached and we're checking for it in `Events/Coupons/CouponCreated.php`. Is it attached implicitly?

tests/Models/CouponTest.php
```diff
@@ -0,0 +1,230 @@
...
+		Config::set('coupons.coupons_sync_token', $expectedToken);
+		Config::set('coupons.coupons_sync_host', $expectedHost);
+		Config::set('coupons.coupons_sync_source', true);
+
+		$mocked->withArgs([
+			'POST',
+			$expectedUrl . "/api/v1/coupons",
+			[
+				'Accept' => 'application/json',
+				'Host' => $expectedToken,
```
rogatty: Doesn't seem right.

app/Events/Coupons/CouponCreated.php
```diff
@@ -0,0 +1,40 @@
...
+
+	public $coupon;
+
+	public function __construct(Coupon $coupon) {
+		$this->method = 'POST';
+		$this->coupon = $coupon;
+
+		$couponToCreate = $this->coupon->toArray();
+		unset($couponToCreate['id']);
+		$this->couponToUpdate = $couponToCreate;
```
rogatty: These three lines are obsolete, I think.

app/Http/Controllers/Payment/PersonalDataController.php
bognix: Yes it is. It checks if the code is the same as one of the codes in StuddyBuddy model.

tests/Models/CouponTest.php
bognix: Could you elaborate?

bognix: The question is why it passed :D

app/Events/Coupons/CouponUpdated.php
bognix: I would prefer to send everything. In case of failure we have more info in request logs.

app/Http/Controllers/Api/PublicApi/CouponsApiController.php
bognix: We will revert, we have all the coupon data in the log.

rogatty: Adam said that our top priority is the codes that should work are working. If we delete incorrect voucher it might be not true. I'd say don't delete when there is more than one and log it.

bognix: I updated the code and added the test case 👍

tests/Api/ApiTestCase.php
```diff
@@ -4,12 +4,19 @@
...
+use Illuminate\Support\Facades\Mail;
 use Tests\TestCase;
 
 abstract class ApiTestCase extends TestCase
 {
 	const BASE = '/papi/v2';
 
+	protected function setUp()
+	{
+		parent::setUp(); // TODO: Change the autogenerated stub
```
bognix: remove the autogenerated TODO comment.

app/Traits/SyncCoupons.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace App\Traits;
+
+use Facades\App\Contracts\Requests;
+
+trait SyncCoupons {
```
bognix: Actually this doesn't have to be a trait. The CouponEvent interface can be an abstract class and we can define this method in the abstract class.

app/Events/Coupons/CouponDeleted.php
```diff
@@ -0,0 +1,25 @@
...
+class CouponDeleted implements CouponEvent {
+	use SerializesModels, SyncCoupons;
+
+	public $coupon;
+
+	public function __construct(Coupon $coupon) {
+		$this->coupon = $coupon;
+	}
+
+	public function shouldSync() {
```
kvas-damian: this method is duplicated, can we have it in one place?

bognix: Yes we can, not sure if it's the best option but I'm flexible and can update the code.

app/Traits/SyncCoupons.php
```diff
@@ -0,0 +1,22 @@
...
+	protected function issueSyncRequest($method, $coupon) {
+		$headers = [
+			'Accept' => 'application/json',
+			'Host' => config('coupons.coupons_sync_host'),
+			config('coupons.coupons_sync_header') => config('coupons.coupons_sync_token'),
+		];
+
+		$url = config('coupons.coupons_sync_url') . "/api/v1/coupons";
```
rogatty: Didn't we deprecate v1?

app/Events/Coupons/CouponDeleted.php
```diff
@@ -0,0 +1,25 @@
...
+		return !empty(config('coupons.coupons_sync_source')) && empty($this->coupon->studyBuddy);
+	}
+
+	public function sync() {
+		$this->issueSyncRequest('DELETE', $this->coupon->toArray());
```
kvas-damian: if this request fails we should revert the delete on the DB.

config/coupons.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+return [
+	'coupons_sync_url' => env('APP_COUPONS_SYNC_URL'),
+	'coupons_sync_host' => env('APP_COUPONS_SYNC_HOST'),
+	'coupons_sync_token' => env('APP_COUPONS_SYNC_TOKEN'),
+	'coupons_sync_source' => env('APP_COUPONS_SYNC_SOURCE'),
```
kvas-damian: `coupons_sync_is_source`

config/coupons.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+return [
+	'coupons_sync_url' => env('APP_COUPONS_SYNC_URL'),
```
kvas-damian: Do we need this? Maybe we can build it based on `coupons_sync_host`

config/papi.php
```diff
@@ -56,6 +56,6 @@
 		'discussions'             => 'discussions',
 		'user-flashcard-notes'    => 'user_flashcard_notes',
 		'roles'                   => 'roles',
-		'course-structure-nodes'  => 'course_structure_nodes',
+		'course-structure-nodes'  => 'course_structure_nodes'
```
kvas-damian: Please don't remove trailing commas :)

app/Events/Coupons/CouponDeleted.php
bognix: We don't have coupon deletion right now. I updated the ticket about coupons CRUD to make the decision then.

app/Traits/SyncCoupons.php
bognix: we depracate `v1` in papi, it's *api*

config/papi.php
bognix: Accident.

.env.travis
```diff
@@ -8,4 +8,4 @@ DB_DATABASE=test_db
 DB_USERNAME=root
 DB_PASSWORD=
 
-MAIL_DRIVER=log
\ No newline at end of file
+MAIL_DRIVER=
```
rogatty: Please use `MAIL_DRIVER=array`, I get an error when using empty value:
```
ArgumentCountError: Too few arguments to function Illuminate\Support\Manager::createDriver(), 0 passed in /code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Support/Manager.php on line 96 and exactly 1 expected

/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Support/Manager.php:85
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Support/Manager.php:96
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Support/Manager.php:71
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Mail/MailServiceProvider.php:102
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Container/Container.php:776
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Container/Container.php:658
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Container/Container.php:609
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Foundation/Application.php:759
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Container/Container.php:1222
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Mail/MailServiceProvider.php:47
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Container/Container.php:776
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Container/Container.php:658
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Container/Container.php:609
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Foundation/Application.php:759
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Container/Container.php:1222
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Support/Facades/Facade.php:175
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Support/Facades/Facade.php:144
/code/bethink/wnl-platform/vendor/laravel/framework/src/Illuminate/Support/Facades/Facade.php:231
/code/bethink/wnl-platform/app/Observers/UserObserver.php:19
```

PR #1312, opened 2019-02-22
PLAT-1019 | Scrape ldek questions

kvas-damian:


app/Models/QuizQuestion.php
```diff
@@ -19,10 +19,6 @@ class QuizQuestion extends Model implements WithReactions, WithTags
 		'preserve_order' => 'boolean',
 	];
 
-	public function setPreserveOrderAttribute($value) {
```
kvas-damian: Because of that, we can't change the value of this field. It's always `false` - you can check on prod.

app/Console/Commands/ScrapeLdekQuizQuestions.php
```diff
@@ -0,0 +1,156 @@
...
+		$cemTaxonomy = Taxonomy::firstOrCreate(['name' => 'Klasyfikacja CEM']);
+		$examTaxonomy = Taxonomy::firstOrCreate(['name' => 'Egzaminy']);
+
+		$bar = $this->output->createProgressBar(count($questions));
+
+		foreach($questions as $question) {
+			$questionCrawler = new Crawler($question);
+			$questionId = $question->getAttribute('pytanieid');
+
+			preg_match("/var pytanieId = '$questionId'.*\\n.*?= '(.)/", $examHtml, $matches);
```
rogatty: Can you add an example of a string we're matching here? Will be easier to debug if we need to update it in the future.

app/Console/Commands/ScrapeLdekQuizQuestions.php
```diff
@@ -0,0 +1,156 @@
...
+			$correctAnswer = $matches[1];
+
+			$questionText = $questionCrawler->filter('.question-and-answear')->first()->html();
```
rogatty: answear 😆

app/Console/Commands/ScrapeLdekQuizQuestions.php
```diff
@@ -0,0 +1,163 @@
...
+			$questionId = $question->getAttribute('pytanieid');
+
+
+			/*
+			 * Match correct answer, example source text
+			 *
+			 * 		var pytanieId = '797';
+			 * 		var odpowiedzPoprawna = 'B.'.replace(" ", "");
+			 */
+			preg_match("/var pytanieId = '$questionId'.*\\n.*?= '(.)/", $examHtml, $matches);
```
rogatty:
```suggestion
			preg_match("/var pytanieId = '$questionId'.*\\n.*?= '([A-E])/", $examHtml, $matches);
```

rogatty: This `.replace(" ", "")` is suspicious, what if they have it because there can be a space at the beginning? Our pattern will fail then:
```
var odpowiedzPoprawna = ' B.'.replace(" ", "");
```

app/Models/QuizQuestion.php
rogatty: We have casting above, anyway.

app/Console/Commands/ScrapeLdekQuizQuestions.php
kvas-damian: Space happens from time to time, but at the end of the string.

We cannot simply use the `[A-E]` matcher because sometimes it's `X` or `?` which means they don't know the answer.

rogatty: Ah, let's add a comment about `X or ?`, then.

kvas-damian:
We can always remove trash when that happens

bognix:
I second @rogatty suggestion. Creating such check is quick. We already had such problems in the past.

PR #1318, opened 2019-02-22
PLAT-1046 | Run Dusk tests in a docker container

rogatty:
https://bethink.atlassian.net/browse/PLAT-1046

To make payment tests almost pass:
- `yarn run dev`
- update columns in `products` table:
  - `quantity` > 0 (each signup decrements it so make it big enough)
  - `access_end` in the future
  - `signups_close` in the future
- add row to `payment_method_product` table with:
  - your `product_id`
  - `payment_method_id = 3` (instalments)
  - `end_date` in the future
- set `APP_URL=http://nginx` in `.env`
- `docker exec -it php /bin/sh`
- `php artisan dusk tests/Browser/Tests/Payment/PaymentTest.php`

You can see what Selenium is doing by going to `vnc://127.0.0.1:5900`. This won't work if we use `selenium/standalone-chrome` (without `-debug`) or `--headless`.


tests/Browser/Tests/Course/Lesson/NavigateUsingSideNavTest.php
```diff
@@ -12,10 +15,22 @@ class NavigateUsingSideNavTest extends DuskTestCase
 
 	private $user;
 
-	public function setUp()
+	public function setUp(): void
 	{
 		parent::setUp();
 		$this->user = factory(User::class)->create();
+		// TODO figure out how to make it cleaner
```
kvas-damian: Do we have a ticket for this TODO? Can we link it?

docker-compose.yaml
```diff
@@ -122,6 +123,17 @@ services:
...
 
+  selenium:
+    container_name: selenium
+    image: selenium/standalone-chrome-debug
+    volumes:
+      - /dev/shm:/dev/shm
+    ports:
+      - "4444:4444"
+      - "5900:5900"
+    links:
```
bognix: Why do we need to set this? I think it should automatically be connected to the network created by docker-compose. The links option is deprecated and may be removed in the future - https://docs.docker.com/compose/compose-file/#links

tests/Browser/Tests/Course/Chat/CourseChatTest.php
```diff
@@ -21,7 +21,7 @@ class CourseChatTest extends DuskTestCase
 	private $user;
 	private $user2;
 
-	public function setUp() {
+	public function setUp(): void {
```
bognix: is it something that we are going to do across the repo? I see previously the php docs were used to indicate the void return type

tests/Browser/Tests/Course/Lesson/NavigateUsingSideNavTest.php
```diff
@@ -12,10 +15,22 @@ class NavigateUsingSideNavTest extends DuskTestCase
...
+		UserSubscription::create([
```
bognix: Can't you use factory here and set up all the required data using factory? This way you won't need to update your own database to start testing.

tests/Browser/Tests/Payment/PaymentRandomTest.php
```diff
@@ -9,6 +9,7 @@ class PaymentRandomTest extends DuskTestCase
 	/** @test */
 	public function randomCheckoutTest()
 	{
+		$this->markTestSkipped();
```
bognix: could you add some comment explaining why this test is skipped.

tests/Browser/Tests/Course/Chat/CourseChatTest.php
rogatty: This method overrides a method from `Laravel\Dusk\TestCase` and the types must be compatible. The original signature was probably changed when we updated Laravel.

But yes, it would be great to add typing wherever possible.

docker-compose.yaml
rogatty: Ah, I took it from some (apparently outdated) article. I confirmed that we can remove it and use `APP_URL=http://nginx` 👍

tests/Browser/Tests/Payment/PaymentRandomTest.php
rogatty: It's something used by @gitkuba and it can't pass because of missing `scenario.dusk` file. Anyway, I don't want to mess with it now, so I reverted this change.

tests/Browser/Tests/Course/Lesson/NavigateUsingSideNavTest.php
rogatty: If I'm correct, you'll need to update it anyway because of the changes needed in `products` and `payment_method_product` tables. I reverted this change because this test is failing anyway. It was here only to show what data is missing. I'll move it to the ticket.

tests/Browser/Tests/Course/Lesson/NavigateUsingSideNavTest.php
rogatty: See https://github.com/bethinkpl/wnl-platform/pull/1318#discussion_r260245657

PR #1320, opened 2019-02-25
[PLAT-1054] Unblock final certificates

preziu:
Users that have finished course after 15/02 can not download final certificate - but it could be possible fim them to do so until the end of course access.


app/Http/Controllers/Api/PrivateApi/CertificatesApiController.php
```diff
@@ -117,13 +117,6 @@ public function getFinalCertificate($orderId)
 			return $this->respondForbidden("User not allowed to view order details");
 		}
 
-		if (!$user->hasFinishedCourse(
```
kvas-damian: Why do we remove this condition? Don't we need to validate if the user finished the course before we generate a certificate?

resources/assets/js/components/user/UserCertificates.vue
```diff
@@ -25,7 +25,7 @@
 			<ul>
 				<li v-for="order in finalCertificates" :key="order.id">
 					<a @click="downloadFinalCertificate(order.id)">
-						Certyfikat Ukończenia: {{order.product.name}} - {{formatDate(order.product.course_start)}} - {{formatDate(order.product.course_end)}}
+						Certyfikat Ukończenia: {{order.product.name}} - {{formatDate(order.product.course_start)}} - {{formatDate(new Date())}}
```
kvas-damian: Are you sure we want to show the current date for courses that finished a year ago?

app/Http/Controllers/Api/PrivateApi/CertificatesApiController.php
```diff
@@ -136,7 +129,7 @@ public function getFinalCertificate($orderId)
 		$img->text(
 			sprintf("%s - %s r.",
 				$order->product->course_start->format('j.m.Y'),
-				$order->product->course_end->format('j.m.Y')
+				$order->product->access_end->format('j.m.Y')
```
kvas-damian: I don't think it's ok if we sign a certificate with future date.

app/Http/Controllers/Api/PrivateApi/CertificatesApiController.php
```diff
@@ -150,7 +143,7 @@ public function getFinalCertificate($orderId)
 			$font->align('center');
 		});
 
-			$img->text($order->product->course_end->format('j / m / Y') . ' r., Poznań', 1182, 867, function($font) {
+			$img->text($order->product->access_end->format('j / m / Y') . ' r., Poznań', 1182, 867, function($font) {
```
kvas-damian: I don't think it's ok that we sign a certificate with a future date.

app/Http/Controllers/Api/PrivateApi/CertificatesApiController.php
```diff
@@ -87,7 +87,7 @@ public function getParticipationCertificate($orderId)
 			$font->align('center');
 		});
 
-		$img->text($order->product->course_start->format('j / m / Y') . ' r., Poznań', 2365, 1650, function($font) {
+		$img->text(Carbon::now()->format('j / m / Y') . ' r., Poznań', 2365, 1650, function($font) {
```
kvas-damian: The date on the certificate will be different on every request. I'd be confused when I get it with different data.

app/Http/Controllers/Api/PrivateApi/CertificatesApiController.php
preziu: The idea behind letting a user download a final certificate is that he can do it after he meets a certain criteria during the course. There is no specific date to be achieved, only course progress conditions are valid. For example: a user met the final criteria at 23.02, but his actual access to the course ends at 1.05. So when we restrict users to the course_end, which at the moment is 15.02, we are closing the opportunity for them to get the final certificate :)

app/Http/Controllers/Api/PrivateApi/CertificatesApiController.php
preziu: My bad! this should stay as a `course_start`, I'll push a fix right away 👍

app/Http/Controllers/Api/PrivateApi/CertificatesApiController.php
preziu: It a certain type of promise. The desired behavior is that we put a date of actual meeting the criteria. This is only a quick fix until we include a proper logic for calculating the actual date.

resources/assets/js/components/user/UserCertificates.vue
preziu: Well, I couldn't really think about a proper date to set the finishing date, but after a thought i think it should be a `access_end`. I'm pushing a fix right away!

app/Http/Controllers/Api/PrivateApi/CertificatesApiController.php
kvas-damian: But right now we don't check if "course progress conditions are valid". Now, everyone can hit this endpoint and generate a certificate for any of his orders - without any checks.

app/Http/Controllers/Api/PrivateApi/CertificatesApiController.php
kvas-damian: Please create a ticket and mark this line with TODO comment.

PR #1324, opened 2019-02-26
PLAT-1039 | Fix is-loading CSS class

kvas-damian:


resources/assets/sass/_styleguide.scss
```diff
@@ -575,8 +575,8 @@ ol {
 	text-transform: uppercase;
 }
 
-
-.is-loading {
+// Make sure our loader doesn't clash with bulma loader
+.is-loading:not(.button, .select, .control) {
```
bognix: in order to make sure to not clash it with bulma I would create a dedicated class name and won't add not selector. What if we want to add our loader to a .button, .select or .control?

PR #1329, opened 2019-02-28
PLAT-1040 | Add laravel schedule monitoring for LDEK

kvas-damian:
Don't merge without: https://github.com/bethinkpl/ansible/pull/52


app/Console/Commands/CommandWithMonitoring.php
```diff
@@ -0,0 +1,29 @@
...
+			if ($metricName) {
+				(new PrometheusPushgateway())->notify(env('MONITORING_METRIC_NAME'));
+			} else {
+				\Log::debug('Monitoring notification silenced on non-production env');
+			}
+		}
+		return $result;
+	}
+
+	abstract public function handleBody();
```
bognix: handleBody sounds odd a little bit. How about execute?

app/Console/Kernel.php
```diff
@@ -33,7 +33,7 @@ protected function schedule(Schedule $schedule)
 			->hourly()
 			->withoutOverlapping()
 			->after(function () {
-				$this->notifyPrometheusPushgateway('orders_stats_export');
+				(new PrometheusPushgateway())->notify('orders_stats_export');
```
bognix: Does it mean that if I run one of those commands using `php artisan...` they will report to Prometheus? Maybe we can add some flag to disable reporting in such case? I think the reporting shouldn't be polluted with "manual" executions if not stated explicitly.

app/Console/PrometheusPushgateway.php
```diff
@@ -0,0 +1,30 @@
...
+namespace App\Console;
+
+
+use GuzzleHttp\Exception\GuzzleException;
+
+class PrometheusPushgateway
+{
+	public function notify($metricName)
+	{
+		$client = new \GuzzleHttp\Client();
```
bognix: could you keep the `use` consistent? Either declare `use ...` at the top of the file or use the absolute names.

config/app.php
```diff
@@ -54,6 +54,16 @@
 
 	'url' => env('APP_URL', 'http://localhost'),
 
+	/*
+	|--------------------------------------------------------------------------
+	| Application Instance Name
+	|--------------------------------------------------------------------------
+	|
+	| Identify different proction environments like lek, ldek, etc...
```
rogatty: `proction`?

app/Console/Commands/StoreProgress.php
```diff
@@ -6,11 +6,10 @@
 use App\Models\UserCourseProgress;
 use Closure;
 use Exception;
-use Illuminate\Console\Command;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Redis;
 
-class StoreProgress extends Command
+class StoreProgress extends CommandWithMonitoring
```
rogatty: Remember to update the rest of the commands too.

app/Console/Kernel.php
rogatty: If you run a command using artisan, you don't go through this flow. You go through `CommandWithMonitoring::handle()` which has a check for `MONITORING_METRIC_NAME` which is set only in k8s cron jobs.

This here is a temporary solution until we migrate LEK production to k8s. Then, we can remove all these Laravel schedule jobs.

app/Console/Commands/StoreProgress.php
kvas-damian: Skleroza...

app/Console/Commands/CommandWithMonitoring.php
kvas-damian: Execute is already reserved: https://github.com/laravel/framework/blob/5.8/src/Illuminate/Console/Command.php#L181
Do you have other names in your mind?

rogatty: `run`

kvas-damian: https://github.com/laravel/framework/blob/5.8/src/Illuminate/Console/Command.php#L163

kvas-damian:
LGTM

PR #1331, opened 2019-03-01
PLAT-1056 | Fix payment browser tests

rogatty:
https://bethink.atlassian.net/browse/PLAT-1056

To run the tests locally use `./tests/payment-tests.sh`.

If they fail, try:
- `yarn run dev`
- update columns in `products` table for the latest `wnl-online` and `wnl-online-onsite` products:
  - `quantity` > 0 (each signup decrements it so make it big enough)
  - `access_end` in the future
  - `signups_close` in the future
- add row to `payment_method_product` table with:
  - `product_id` of the latest `wnl-online` product
  - `payment_method_id = 3` (instalments)
  - `end_date` in the future


.gitignore
```diff
@@ -5,6 +5,9 @@
 /.vscode
 .env
 .env.backup
+.env.dusk.*
+.env.ngrok
+.env.selenium
```
bognix: is there any `.env` file that we want to commit?

tests/payment-tests.sh
```diff
@@ -0,0 +1,24 @@
...
+cd ${DIR}/..
+
+NGROK_URL=$(./ngrok-enable.sh -r 0)
+
+# ngrok failed
+if [[ $? -eq 1 ]]; then
+    exit 1
+fi
+
+P24_STATUS_URL="$NGROK_URL/payment/status" APP_URL="http://nginx" SESSION_DOMAIN="nginx" docker-compose up -d php
```
bognix: perhaps `docker restart php` would work as well?

rogatty: I'm not sure if it'll work because it has to reload environment variables listed in `docker-compose.yaml`. Is there a good reason to check it? Is `docker restart php` better?

tests/payment-tests.sh
```diff
@@ -0,0 +1,26 @@
...
+cd ${DIR}/..
+
+NGROK_URL=$(./ngrok-enable.sh -r 0)
+
+# ngrok failed
+if [[ $? -eq 1 ]]; then
+    exit 1
+fi
+
+P24_STATUS_URL="$NGROK_URL/payment/status" APP_URL="http://nginx" SESSION_DOMAIN="nginx" DEBUG_BAR="false" docker-compose up -d php
```
kvas-damian: `DEBUG_BAR` generates code that makes selectors not matching correct elements - so I disabled it.

tests/payment-tests.sh
bognix: It was more explicit about what we do here but I'm not sure if it will re-read the variable for docker-compose. I highly doubt it.

PR #1332, opened 2019-03-04
[PLAT-997] Admin can CRU a product

gitkuba:


resources/assets/js/admin/components/products/ProductEdit.vue
```diff
@@ -0,0 +1,120 @@
+<template>
+	<div>
+		<wnl-form
+				name="DashboardNewsEditor"
```
rogatty: Doesn't seem right.

resources/assets/js/admin/components/products/ProductEdit.vue
```diff
@@ -0,0 +1,120 @@
...
+		>
+			<h4>Produkt #{{ id }}</h4>
+			<wnl-text name="name">Nazwa</wnl-text>
+			<wnl-text name="slug">Slug</wnl-text>
+			<wnl-text name="price">Cena</wnl-text>
+			<wnl-select name="vat_rate" :options="vatRates">Stawka VAT</wnl-select>
+			<wnl-text name="vat_note">Opis VAT</wnl-text>
+			<wnl-textarea name="invoice_name">Opis na fakturze</wnl-textarea>
+
+			<wnl-text name="quantity">Ilość (obecnie)</wnl-text>
```
rogatty: Ilość wody, liczba produktów.

resources/assets/js/admin/components/products/ProductEdit.vue
```diff
@@ -0,0 +1,120 @@
...
+				altInput: true,
+				enableTime: true,
+				dateFormat: 'U',
+				altFormat: 'Y-m-d H:i',
+				time_24hr: true,
+			},
+			formData: {}
+		};
+	},
+	props: ['id'],
```
rogatty: Let's validate it.

resources/assets/js/admin/components/products/ProductsList.vue
```diff
@@ -0,0 +1,79 @@
...
+		<h3 class="title is-3">
+			Produkty
+			<router-link
+					class="button is-primary"
+					:to="{ name: 'products-edit', params: { id: 'new' } }"
+			>
+				+ Dodaj nowy produkt
+			</router-link>
+		</h3>
+		<table class="table dashboard-news">
```
rogatty: Doesn't seem right.

resources/assets/js/admin/components/products/ProductsList.vue
```diff
@@ -0,0 +1,79 @@
...
+			<tr>
+				<th>Id</th>
+				<th>Tytuł</th>
+				<th>Dostępność</th>
+				<th>Start zapisów</th>
+				<th>Utworzono</th>
+			</tr>
+			<tr
+					class="dashboard-news__item"
```
rogatty: Doesn't seem right.

resources/assets/js/admin/components/products/ProductsList.vue
```diff
@@ -0,0 +1,79 @@
...
+</style>
+
+<script>
+import { mapActions } from 'vuex';
+import moment from 'moment';
+import {getApiUrl} from 'js/utils/env';
+import {ALERT_TYPES} from 'js/consts/alert';
+
+export default {
+	name: 'DashboardNews',
```
rogatty: Doesn't seem right.

resources/assets/js/admin/components/products/ProductsList.vue
```diff
@@ -0,0 +1,79 @@
...
+			}
+		},
+		goToEdit(id) {
+			this.$router.push({ name: 'products-edit', params: { id } });
+		},
+	},
+	async mounted() {
+		try {
+			const {data} = await axios.get(getApiUrl('products/all'));
+			this.products = Object.values(data).sort((a, b) => a.id < b.id ? 1 : -1);
```
rogatty: Can we sort them on the backend?

resources/assets/js/admin/router.js
```diff
@@ -185,6 +187,17 @@ let routes = [
...
 	},
+	{
+		name: 'products',
+		path: '/admin/app/products',
+		component: ProductsList,
+	},
+	{
+		name: 'products-edit',
+		path: '/admin/app/products/:id',
+		component: ProductEdit,
```
rogatty: `name: 'products-edit'` but `component: ProductEdit`, I'd pick one.

app/Http/Controllers/Api/PrivateApi/ProductsApiController.php
```diff
@@ -0,0 +1,77 @@
...
+			'delivery_date' => Carbon::createFromTimestamp($request->delivery_date),
+			'course_start'  => Carbon::createFromTimestamp($request->course_start),
+			'course_end'    => Carbon::createFromTimestamp($request->course_end),
+			'access_start'  => Carbon::createFromTimestamp($request->access_start),
+			'access_end'    => Carbon::createFromTimestamp($request->access_end),
+			'signups_start' => Carbon::createFromTimestamp($request->signups_start),
+			'signups_end'   => Carbon::createFromTimestamp($request->signups_end),
+			'signups_close' => Carbon::createFromTimestamp($request->signups_close),
+			'vat_rate'      => $request->vat_rate,
+			'vat_note'      => $request->vat_note,
```
kvas-damian: Let's create a helper method to transform request data into a model format to not duplicate it in post and put.

app/Models/Product.php
```diff
@@ -6,9 +6,25 @@
 
 class Product extends Model
 {
+	const VAT_RATES = [0, 5, 8 , 23];
```
kvas-damian: What if we need to update vat rates? By hardcoding them here, we need to perform a release. Did you consider putting them in the DB?

app/Http/Controllers/Api/Transformers/ProductTransformer.php
```diff
@@ -0,0 +1,51 @@
...
+use App\Http\Controllers\Api\ApiTransformer;
+use App\Models\Product;
+
+class ProductTransformer extends ApiTransformer
+{
+
+	protected $availableIncludes = [];
+	protected $parent;
+
+	public function __construct($parent = null)
```
kvas-damian: remove empty contructor

app/Http/Requests/Product/CreateProduct.php
```diff
@@ -0,0 +1,47 @@
...
+
+	/**
+	 * Get the validation rules that apply to the request.
+	 *
+	 * @return array
+	 */
+	public function rules()
+	{
+		return [
+			'id'            => 'required|integer|exists:products,id',
```
kvas-damian: `id` is required for product creation?

app/Http/Controllers/Api/PrivateApi/ProductsApiController.php
```diff
@@ -0,0 +1,77 @@
...
+		]);
+
+		$this->transformAndRespond($product);
+	}
+
+	public function getVatRates()
+	{
+		return $this->respondOk(['vat_rates' => Product::VAT_RATES]);
```
kvas-damian: technically vat rates are not part of a product resource.

resources/assets/js/admin/components/Sidenav.vue
```diff
@@ -54,6 +54,11 @@
 		<ul class="menu-list">
 			<li><router-link :to="{ name: 'orders' }">Lista zamówień</router-link></li>
 		</ul>
+		<p class="menu-label">Produkty</p>
+		<ul class="menu-list">
+			<li><router-link :to="{ name: 'product-edit', params: { id: 'new' } }">Dodaj produkt</router-link></li>
```
kvas-damian: We follow a convention to put a create link on the listing page. Do we want to change it everywhere?

resources/assets/js/admin/components/products/ProductEdit.vue
```diff
@@ -0,0 +1,149 @@
...
+					<hr>
+
+					<wnl-datepicker name="access_start" :config="datepickerConfig">Początek dostępu</wnl-datepicker>
+					<wnl-datepicker name="access_end" :config="datepickerConfig">Koniec dostępu</wnl-datepicker>
+
+					<hr>
+
+					<wnl-datepicker name="signups_start" :config="datepickerConfig">Start zapisów</wnl-datepicker>
+					<wnl-datepicker name="signups_end" :config="datepickerConfig">Koniec zapisów</wnl-datepicker>
+					<wnl-datepicker name="signups_close" :config="datepickerConfig">Zamknięcie zapisów</wnl-datepicker>
```
kvas-damian: We should have an info box explaining the difference between `signups_end` and `signups_close`. It's not clear for me, so probably end user will be even more confused.

resources/assets/js/admin/components/products/ProductEdit.vue
```diff
@@ -0,0 +1,149 @@
...
+		},
+	},
+	methods: {
+		...mapActions(['addAutoDismissableAlert']),
+		onSubmitSucess(data) {
+			if (!this.isEdit) {
+				this.$router.push({name: 'product-edit', params: {id: data.id}});
+			}
+		},
+		onChange({formData}) {
```
kvas-damian: You can remove it

resources/assets/js/admin/components/products/ProductEdit.vue
```diff
@@ -0,0 +1,149 @@
...
+			this.formData = formData;
+		},
+		escapeArgumentKey(key) {
+			return `{{${key}}}`;
+		},
+	},
+	async mounted() {
+		try {
+			const {data} = await axios.get(getApiUrl('products/getVatRates'));
+			this.vatRates = data.vat_rates.map(value => {
```
kvas-damian: `data.vat_rates.map(value => ({value, text: value})`

resources/assets/js/admin/components/products/ProductsList.vue
```diff
@@ -0,0 +1,91 @@
...
+			<router-link
+					class="button is-primary"
+					:to="{ name: 'product-edit', params: { id: 'new' } }"
+			>
+				+ Dodaj nowy produkt
+			</router-link>
+		</h3>
+
+		<wnl-paginated-sortable-table
+				:is-search-enabled="false"
```
kvas-damian: why do we disable search?

resources/assets/js/admin/components/products/ProductsList.vue
```diff
@@ -0,0 +1,91 @@
...
+				:resource-name="'products/.filter'"
+				:columns="columns"
+				class="products"
+		>
+			<tbody slot-scope="slotProps" slot="tbody">
+			<tr v-for="product in slotProps.list" :key="product.id" @click="goToEdit(product.id)" class="products__item">
```
kvas-damian: `class=clickable`

resources/assets/js/admin/components/products/ProductsList.vue
```diff
@@ -0,0 +1,91 @@
...
+				{
+					name: 'signups_start',
+					label: 'Start zapisów',
+				},
+				{
+					name: 'created_at',
+					label: 'Utworzono',
+				},
+			],
+			requestParams:{
```
kvas-damian: not used

resources/assets/js/admin/components/products/ProductsList.vue
```diff
@@ -0,0 +1,91 @@
...
+		};
+	},
+	components: {
+		WnlPaginatedSortableTable,
+	},
+	methods: {
+		...mapActions(['addAutoDismissableAlert']),
+		formatDate(date) {
+			if (date) {
+				return moment(date * 1000).format('DD/MM/YY H:mm');
```
kvas-damian: I don't like such custom date formats. Moment provides a set of standard formats for every local that we can use. In this case I think `L LT` is the right format to use.

app/Http/Controllers/Api/PrivateApi/ProductsApiController.php
gitkuba: Seems like an overkill to me. At least for now.

app/Models/Product.php
gitkuba: I don't feel like refactoring invoice generator. Let's wait for more use cases before implementing vat rate as a new entity.

resources/assets/js/admin/components/products/ProductsList.vue
gitkuba: What's the use case for search here?

resources/assets/js/admin/components/products/ProductEdit.vue
```diff
@@ -0,0 +1,144 @@
...
+	data() {
+		return {
+			datepickerConfig: {
+				altInput: true,
+				enableTime: true,
+				dateFormat: 'U',
+				altFormat: 'Y-m-d H:i',
+				time_24hr: true,
+			},
+			formData: {},
```
kvas-damian: unused

resources/assets/js/admin/components/products/ProductsList.vue
kvas-damian: Probably to search by name if there are more products in the DB. Not sure if it's a valid use-case now.

gitkuba: not sure as well

resources/assets/js/admin/components/products/ProductEdit.vue
```diff
@@ -0,0 +1,155 @@
...
+					<hr>
+
+					<wnl-datepicker name="delivery_date" :config="datepickerConfig">Data dostawy</wnl-datepicker>
+				</div>
+			</div>
+		</wnl-form>
+
+		<article class="message is-primary">
+			<div class="message-header">
+				<p>Legenda</p>
```
kvas-damian: I have a feeling that legend isn't the best UX. IMO, we should keep such information as close to the input field as possible. What do you think?

app/Http/Controllers/Api/PrivateApi/ProductsApiController.php
```diff
@@ -0,0 +1,64 @@
...
+		$productId = $request->get('id');
+		$product = Product::find($productId);
+
+		if (is_null($product)) {
+			return $this->respondNotFound();
+		}
+
+		$product->update($this->transformRequestParams($request));
+
+		return $this->respondOk();
```
bognix: How about returning updated product? You return created product when you issue post request.

app/Http/Controllers/Api/PrivateApi/ProductsApiController.php
```diff
@@ -0,0 +1,64 @@
...
+
+	private function transformRequestParams($request) {
+		return [
+			'name'          => $request->name,
+			'invoice_name'  => $request->invoice_name,
+			'slug'          => $request->slug,
+			'price'         => $request->price,
+			'quantity'      => $request->quantity,
+			'initial'       => $request->initial,
+			'delivery_date' => Carbon::createFromTimestamp($request->delivery_date),
```
bognix: isn't it possible to send data from frontend in the right shape so you won't have to call this Carbon method on each field?

gitkuba: Yup, entirely posibble. Does it make a difference?

resources/assets/js/admin/components/products/ProductEdit.vue
```diff
@@ -0,0 +1,155 @@
...
+		},
+	},
+	methods: {
+		...mapActions(['addAutoDismissableAlert']),
+		onSubmitSucess(data) {
+			if (!this.isEdit) {
+				this.$router.push({name: 'product-edit', params: {id: data.id}});
+			}
+		},
+		escapeArgumentKey(key) {
```
bognix: I think this is not used.

app/Http/Controllers/Api/PrivateApi/ProductsApiController.php
bognix: Yes, you don't need this method then, just call `$request->all()`

app/Http/Controllers/Api/PrivateApi/ProductsApiController.php
```diff
@@ -0,0 +1,64 @@
+<?php namespace App\Http\Controllers\Api\PrivateApi;
+
+use App\Http\Controllers\Api\ApiController;
+use App\Http\Requests\Product\CreateProduct;
+use App\Http\Requests\Product\UpdateProduct;
+use App\Models\Product;
+use Illuminate\Http\Request;
+use Carbon\Carbon;
+
+class ProductsApiController extends ApiController
```
bognix: please add API tests for this controller.

resources/assets/js/admin/components/products/ProductEdit.vue
```diff
@@ -0,0 +1,155 @@
+<template>
+	<div>
+		<wnl-form
+				name="ProductsEditor"
+				:populate="isEdit"
+				:method="formMethod"
+				:resource-route="formResourceRoute"
+				:suppress-enter="true"
+				@submitSuccess="onSubmitSucess"
+				@change="onChange"
```
bognix: onChange method is not defined.

bognix:
In our application we have `ProductsTransformer` - can we get rid of it?

app/Http/Controllers/Api/PrivateApi/ProductsApiController.php
gitkuba: It's good to be explicit about fields that are beeing updated. Passing `$request->all()` directly to `update()` is an anti-pattern imho.

bognix: cc @kvas-damian @rogatty since I saw this pattern used by one of you :)

tests/Api/Products/ProductsTest.php
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	/** @test */
+	public function update_product()
+	{
+		$user = factory(User::class)->create();
+		$user->roles()->attach(Role::byName('admin'));
+		$product = factory(Product::class)->create();
+
+		$updated = $this->transformProduct(factory(Product::class)->create(['slug' => null]));
```
bognix: Why do you change slug to null? You are creating totally different product so everything faker returns should be different.

gitkuba: OK, nvm, i've found a better solution :)

app/Http/Requests/Product/UpdateProduct.php
```diff
@@ -0,0 +1,52 @@
...
+		    'initial'       => 'integer',
+		    'delivery_date' => 'integer',
+		    'course_start'  => 'integer|nullable',
+		    'course_end'    => 'integer|nullable',
+		    'access_start'  => 'integer|nullable',
+		    'access_end'    => 'integer|nullable',
+		    'signups_start' => 'integer',
+		    'signups_end'   => 'integer',
+		    'signups_close' => 'integer',
+		    'vat_rate'      => 'numeric',
```
kvas-damian: are these fields optional? Don't we need to match this validation with `CreateProduct.php`?

PR #1335, opened 2019-03-04
PLAT-992 | A user can access a lesson without a date defined in her plan

rogatty:
https://bethink.atlassian.net/browse/PLAT-992

To do:

- [x] `CalculateCoursePlan::handleDefaultPlan` should remove the custom plan
- [x] Load the start dates in a more Laravelesque way, if possible
- [x] Batch queries for `user_lesson`
- [x] Confirm with Adam that sorting by `product.course_start` is the correct way to decide which product is "the newest bought" one
- [x] Update planner's UI to mark default dates, if needed
- [x] on LEK make lessons in group id `14` required: `update lessons set is_required = 1 where group_id = 14;`


app/Models/Lesson.php
```diff
@@ -50,56 +53,73 @@ public function getQuestionsAttribute()
...
+		return $this->setStartDate(null);
 	}
 
-	public function userLessonAccess(User $user) {
-		return DB::table('user_lesson')
-			->where('lesson_id', $this->id)
-			->where('user_id', $user->id)
-			->first();
+	private function setStartDate($startDate) {
+		$this->startDate = $startDate;
```
gitkuba: I'm not sure why save it to object property? Is it because we're going to call it multiple times during one request?

app/Models/Lesson.php
```diff
@@ -50,56 +53,73 @@ public function getQuestionsAttribute()
...
+	public function startDate()
 	{
-		$user = $user ?? \Auth::user();
-		if ($user) {
-			$lessonAccess = $this->userLessonAccess($user);
+		if ($this->startDateLoaded) {
+			return $this->startDate;
+		}
+
+		$user = \Auth::user();
```
gitkuba: If we decide some day to use this method on queue, `\Auth::user()` won't be available (will return `null` probably).

app/Models/Lesson.php
gitkuba: Also, if i correctly understood `startDate()` method, it's possible, that this date will be valid for a particular user. I think `Lesson` model shouldn't store such information(?)

app/Models/User.php
```diff
@@ -283,6 +286,26 @@ protected function getSubscriptionDates()
 		return [$min, $max];
 	}
 
+	public function getProductIdForDefaultLessonsStartDates() {
```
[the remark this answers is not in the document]
kvas-damian: This name is little confusing. Instead of naming it by purpose maybe we can tell what is actually is?
maybe `getRecentProductId()`?

app/Models/Lesson.php
rogatty: Yes, we're calling this method multiple times. We make 249 DB queries on master and 92 DB queries on this branch, when requesting course structure. I'll try to lower it even more.

app/Models/Lesson.php
```diff
@@ -50,56 +53,73 @@ public function getQuestionsAttribute()
...
-		if ($user) {
-			if ($user->isAdmin() || $user->isModerator()) {
-				return true;
-			}
-
-			$lessonAccess = $this->userLessonAccess($user);
-			if (!is_null($lessonAccess) && !is_null($lessonAccess->start_date)) {
-				return Carbon::parse($lessonAccess->start_date)->isPast();
-			}
+		$user = \Auth::user();
```
bognix: $user can be null. Auth::user in some contexts may return null - right now this will fail with error.

app/Models/Lesson.php
```diff
@@ -50,56 +53,73 @@ public function getQuestionsAttribute()
...
+		if ($user->isAdmin() || $user->isModerator()) {
+			return true;
 		}
 
-		return false;
+		return !is_null($this->startDate());
 	}
 
-	public function startDate($user = null)
+	public function startDate()
```
bognix: the name is not intuitive in my opinion. If you define a setter (setStartDate) I would consider naming this method `getStartDate`.

app/Models/Lesson.php
bognix: @gitkuba I think we all agree that user access shouldn't be stored on the lesson level. However, this was implemented like this long time ago and I'm not sure if this is the best time to decoupling. I think @rogatty's solution is just a small optimization, not a final solution. I thought we already have a ticket to do this decoupling, apparently we don't so let's create one, add a comment above this flag and move on with our lives for now :)

app/Models/User.php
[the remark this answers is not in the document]
bognix: I agree with @kvas-damian it's not returning dates after all, it's returning an id.

app/Models/User.php
```diff
@@ -283,6 +286,26 @@ protected function getSubscriptionDates()
...
+				->where('orders.user_id', $this->id)
+				->where('orders.paid', 1)
+				->orderBy('course_start', 'desc')
+				->first();
+
+			if ($product) {
+				$this->productIdForDefaultLessonsStartDates = $product->id;
+			}
+
+			$this->productIdForDefaultLessonsStartDatesLoaded = true;
```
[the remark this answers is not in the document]
bognix: I think we don't want to do this check again if user doesn't have any bought products.

app/Models/User.php
```diff
@@ -283,6 +286,26 @@ protected function getSubscriptionDates()
 		return [$min, $max];
 	}
 
+	public function getProductIdForDefaultLessonsStartDates() {
+		if (!$this->productIdForDefaultLessonsStartDatesLoaded) {
+			$product = Product::select(['products.id'])
```
bognix: I know that our policy right now is not to cache stuff but maybe it's a good candidate to cache this value in redis and only invalidate when order is paid.

app/Models/User.php
```diff
@@ -283,6 +283,59 @@ protected function getSubscriptionDates()
...
+				$this->productIdForDefaultLessonsStartDates = $product->id;
+			}
+
+			$this->productIdForDefaultLessonsStartDatesLoaded = true;
+		}
+
+		return $this->productIdForDefaultLessonsStartDates;
+	}
+
+	public function getProductLessons()
```
kvas-damian: This name doesn't suggest that we're excluding userLessons.

app/Models/User.php
rogatty: This query is quick enough that I don't think we need the cache:
```
mysql> explain select `products`.`id` from `products` inner join `orders` on `orders`.`product_id` = `products`.`id` inner join `lesson_product` on `lesson_product`.`product_id` = `products`.`id` where `orders`.`user_id` = 3307 and `orders`.`paid` = 1 order by `course_start` desc limit 1;
+----+-------------+----------------+------------+--------+--------------------------------------------------+---------------------------------+---------+-----------------------+------+----------+----------------------------------------------+
| id | select_type | table          | partitions | type   | possible_keys                                    | key                             | key_len | ref                   | rows | filtered | Extra                                        |
+----+-------------+----------------+------------+--------+--------------------------------------------------+---------------------------------+---------+-----------------------+------+----------+----------------------------------------------+
|  1 | SIMPLE      | orders         | NULL       | ref    | orders_user_id_foreign,orders_product_id_foreign | orders_user_id_foreign          | 4       | const                 |    3 |    10.00 | Using where; Using temporary; Using filesort |
|  1 | SIMPLE      | products       | NULL       | eq_ref | PRIMARY                                          | PRIMARY                         | 4       | wnl.orders.product_id |    1 |   100.00 | NULL                                         |
|  1 | SIMPLE      | lesson_product | NULL       | ref    | lesson_product_product_id_index                  | lesson_product_product_id_index | 4       | wnl.orders.product_id |   64 |   100.00 | Using index                                  |
+----+-------------+----------------+------------+--------+--------------------------------------------------+---------------------------------+---------+-----------------------+------+----------+----------------------------------------------+
3 rows in set, 1 warning (0.01 sec)
```

If you wonder about the warning, it's https://dev.mysql.com/doc/refman/5.7/en/explain-extended.html:
```
mysql> show warnings limit 1;
+-------+------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Level | Code | Message                                                                                                                                                                                                                                                                                                                                                                                     |
+-------+------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Note  | 1003 | /* select#1 */ select `wnl`.`products`.`id` AS `id` from `wnl`.`products` join `wnl`.`orders` join `wnl`.`lesson_product` where ((`wnl`.`products`.`id` = `wnl`.`orders`.`product_id`) and (`wnl`.`lesson_product`.`product_id` = `wnl`.`orders`.`product_id`) and (`wnl`.`orders`.`paid` = 1) and (`wnl`.`orders`.`user_id` = 3307)) order by `wnl`.`products`.`course_start` desc limit 1 |
+-------+------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

app/Contracts/CourseProvider.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace App\Contracts;
+
+class CourseProvider {
+
+	static $DEFAULT_COURSE_ID = 1;
+
+	public function getCourseId() {
+		return static::$DEFAULT_COURSE_ID;
```
kvas-damian: "szpachla" under cover :P

app/Console/Commands/StoreProgress.php
```diff
@@ -46,11 +47,12 @@ public function handleCommand()
 	{
 		// IMPORTANT!!!! user id is in fact profile id
 		$passedUserId = $this->argument('user');
+		$courseId = CourseProvider::getCourseId();
```
kvas-damian: you can move it into the transaction

app/Models/Lesson.php
```diff
@@ -44,56 +48,64 @@ public function getQuestionsAttribute()
...
-			->first();
+	/**
+	 * TODO move it away from Lesson model
+	 *
+	 * @param User|null $user
+	 * @return Carbon|null
+	 */
+	public function startDate(User $user = null)
+	{
+		$user = $user ?? Auth::user();
```
bognix: what if `Auth::user`` return null? It will break. Can we handle this situation?

app/Models/User.php
```diff
@@ -160,6 +154,12 @@ public function userTime() {
 		return $this->hasMany('App\Models\UserTime');
 	}
 
+	public function userLessons()
```
bognix: this relation doesn't work correctly now. Because it will only return lessons user has in user_lesson table. Right now it should join lessons from user_lesson and lesson_product

app/Models/User.php
bognix: Ok if the query is quick then why do you use the `productIdForDefaultLessonsStartDatesLoaded` flag?
I was proposing cache as the alternative for the flag because I have an impression that it slightly pollutes models.

app/Models/Lesson.php
rogatty: Handled in https://github.com/bethinkpl/wnl-platform/pull/1335/commits/0a4f29ac69f409562019209a04c827a5cf139c1e

app/Models/User.php
```diff
@@ -283,6 +283,62 @@ protected function getSubscriptionDates()
 		return [$min, $max];
 	}
 
+	public function getLatestPaidCourseProductId() {
```
bognix: could you create unit tests for this method?

app/Models/User.php
rogatty: That's what we do in `getLessonsAvailability`

app/Models/User.php
```diff
@@ -283,6 +283,62 @@ protected function getSubscriptionDates()
...
+			->get()
+			->each(function (Lesson $lesson) {
+				$lesson->is_default_start_date = true;
+			});
+	}
+
+	/**
+	 * @return Collection
+	 */
+	public function getLessonsAvailability()
```
bognix: could you create unit tests for this method?

app/Models/User.php
bognix: ignore my comment.

app/Models/User.php
rogatty: https://github.com/bethinkpl/wnl-platform/commit/1b68f9486e6a8444d4671da3482107d195356ba1

PR #1336, opened 2019-03-05
PLAT-996 | Remove workshops group

kvas-damian:
I decided to run SQL manually to make sure we don't remove something wrong on LDEK using a migration.
SQL is independent from code changes:
```
delete from user_course_progress where lesson_id = 17;
delete from lesson_user_access where lesson_id = 17;
delete from lesson_product where lesson_id = 17;
delete from user_lesson where lesson_id = 17;
delete from screens where lesson_id = 17;

delete from lessons where id = 17;
delete from groups where id = 3;
```

TODO:
* [ ] run SQL on prod and staging;


rogatty:
Can we use PHP to delete the course structure nodes? I don't like that we manually update the tree's internals. At least run https://github.com/lazychaser/laravel-nestedset#fixing-tree after deletion, instead the manual updates.

kvas-damian:
@rogatty I don't feel a need for it. We don't want to perform these operations on LDEK, so the update for lek is very simple. Although, if you feel safer I can add a command for that.

kvas-damian:
We decided to remove these 2 items from the tree using admin panel

PR #1340, opened 2019-03-08
PLAT-1090 | Cleanup after course structure change

kvas-damian:
Removed outdated commands:
* DumpCourseStructure.php
* LessonUser.php


resources/assets/js/store/modules/course.js
```diff
@@ -37,7 +37,20 @@ const getters = {
 	ready: state => state.ready,
 	courseId: state => state.id,
 	name: state => state.name,
+	getNode: state => nodeId => state.structure.find(node => node.id === nodeId),
 	getChildrenNodes: state => parentId => state.structure.filter(node => node.parent_id === parentId),
+	getAncestorNodes: (state, getters) => nodeId => {
```
bognix: we already have something like this defined in nestedSet/getAncestorById - can we unify or create a shared method that can be reused?

app/Console/Commands/DumpCourseStructure.php
```diff
@@ -24,23 +24,16 @@ class DumpCourseStructure extends Command
...
-	}
-
 	/**
 	 * Execute the console command.
 	 *
 	 * @return mixed
 	 */
 	public function handle()
 	{
+		$this->error('This Command isn\'t ready for tree course structure');
```
rogatty: I'd prefer to delete this command than break it like this.

bognix:
@kvas-damian fix seeders.

app/Console/Commands/DumpCourseStructure.php
```diff
@@ -24,23 +24,16 @@ class DumpCourseStructure extends Command
 	 */
```
gitkuba: File can be removed

resources/assets/js/store/modules/course.js
kvas-damian: I have a feeling using shared method will make the data flow harder to follow. We'd need to extract this method somewhere and pass state there. What do you think about using the same name?

app/Http/Controllers/Api/PrivateApi/UserStateApiController.php
```diff
@@ -179,7 +179,13 @@ static function countUserStats(User $user)
 		$qnaAnswersPosted = QnaAnswer::where('user_id', $userId)->count();
 		$quizQuestionsSolved = UserQuizResults::where('user_id', $userId)->groupBy('quiz_question_id')->get(['quiz_question_id'])->count();
 		$numberOfQuizQuestions = QuizQuestion::count();
-		$numberOfLessons = Lesson::whereNotIn('group_id', [3])->count();
+		$numberOfLessons = $user->getLessonsAvailability()->reduce(function (int $count, Lesson $lesson) {
+			if ($lesson->isAvailable()) {
```
bognix: @rogatty I think it should be isAccessible not isAvailable.

rogatty: Agreed 👍

database/migrations/2019_02_01_160121_recreate_course_structure.php
```diff
@@ -1,31 +0,0 @@
...
-	/**
-	 * Run the migrations.
-	 *
-	 * @return void
-	 */
-	public function up()
-	{
-		DB::table('course_structure_nodes')->truncate();
-		// First version of the script messed up the ordering, let's recreate the structure with a fixed script
-		\Artisan::call('data-migration:create-nested-course-structure');
```
rogatty: I'm not sure if it's a good practice to remove a migration as its name is already in the prod DB. Maybe just remove this line?

kvas-damian: This migration doesn't work anymore as we're dropping the groups&lessons relation.

rogatty: A lot of old migrations don't work if you run them on the current state of DB but it's not a reason to remove them all. Quoting https://laracasts.com/discuss/channels/laravel/deleting-a-migration:
>i want to delete a migration file but when i did i cannot migrate:refresh again because it cannot find the deleted file

>Rule one of migration, never delete your migration files if you have migrated them.

kvas-damian: Artisan is still able to perform new migrations, so I don't see any reason to keep empty file...

kvas-damian: I reverted changes and kept empty migration

resources/assets/js/store/modules/course.js
bognix: ok, let's start with that ;)

PR #1342, opened 2019-03-11
PLAT-1064 | Add Cart

bognix:


app/Http/Controllers/Payment/PersonalDataController.php
```diff
@@ -55,9 +55,26 @@ public function index(FormBuilder $formBuilder, $productSlug = null)
...
 
+		$coupon = $this->readCoupon(Auth::user());
+		$productPriceWithCoupon = null;
+		$couponValue = null;
+
+		if (!empty($coupon)) {
+			if ($coupon->is_percentage) {
+				$value = number_format($coupon->value * $product->price / 100, 2, '.', '');
+				$productPriceWithCoupon = $product->price - $value;
+				$couponValue = "- {$coupon->value} %";
```
bognix: the decision about `%` or `zl` could go to view potentially.

app/Http/Controllers/Payment/ConfirmOrderController.php
```diff
@@ -28,13 +28,27 @@ public function index(Payment $payment)
 		$amount = (int)$order->total_with_coupon * 100;
 		$checksum = $payment::generateChecksum($order->session_id, $amount);
 
+		$coupon = $order->coupon;
+		$productPriceWithCoupon = $order->totalWithCoupon;
```
rogatty: Is it different from `$amount = (int)$order->total_with_coupon * 100;` above? If so, let's add a comment explaining the difference.

app/Http/Controllers/Payment/PersonalDataController.php
rogatty: Agreed, formatting belongs to the view layer.

app/Http/Controllers/Payment/ConfirmOrderController.php
bognix: It's the same but not in "grosze". I'll reuse the calculation.

app/Http/Controllers/Payment/PersonalDataController.php
```diff
@@ -55,9 +55,26 @@ public function index(FormBuilder $formBuilder, $productSlug = null)
...
 		session()->flash('url.intended', route('payment-personal-data'));
 
+		$coupon = $this->readCoupon(Auth::user());
+		$productPriceWithCoupon = null;
+		$couponValue = null;
+
+		if (!empty($coupon)) {
+			if ($coupon->is_percentage) {
+				$value = number_format($coupon->value * $product->price / 100, 2, '.', '');
+				$productPriceWithCoupon = $product->price - $value;
```
rogatty: `number_format` returns a string, we should use the numeric version for calculations.

app/Http/Controllers/Payment/PersonalDataController.php
```diff
@@ -112,15 +129,12 @@ protected function createOrder($user, $request)
...
+		$coupon = $this->readCoupon(Auth::user());
 
-		$userCoupon = $user->coupons->first();
-		if (session()->has('coupon')) {
-			$coupon = session()->get('coupon')->fresh();
-			$this->addCoupon($order, $coupon);
-		} elseif ($userCoupon) {
-			$this->addCoupon($order, $userCoupon);
-		} elseif ($order->product->slug !== 'wnl-album') {
+		if (empty($coupon) && $order->product->slug !== 'wnl-album') {
```
rogatty:
```
if (!empty($coupon)) {
	...
} else if ($order->product->slug !== 'wnl-album') {
	...
}
```

resources/views/payment/cart.blade.php
```diff
@@ -0,0 +1,33 @@
+
+<div class="card column cart{{$hasCoupon ? ' has-coupon' : ''}}">
+    @if(!empty($productName))
+    <header class="card-header">
```
rogatty: I think there should be an indent here.

resources/views/payment/cart.blade.php
```diff
@@ -0,0 +1,33 @@
...
+        <h3 class="card-header-title">@lang('payment.cart-header')</h3>
+    </header>
+    <div class="card-content">
+        <p>
+            @lang('payment.cart-access-info', ['date' => $productAccessEnd->format('d.m.Y')])
+        </p>
+        <p>
+            <span>@lang('payment.cart-shipment-label')</span>
+            <span>@lang('payment.cart-shipment-value')</span>
```
rogatty: It would be more semantically correct to use a `dd` list.

resources/views/payment/confirm-order.blade.php
```diff
@@ -1,6 +1,16 @@
 @extends('payment.layout')
 
 @section('content')
+
+	@include('payment.cart', [
+		'productName' => $order->product->name,
+		'productPrice' => $order->product->price,
+		'productAccessEnd' => $order->product->access_end,
+		'hasCoupon' => !empty($couponValue),
```
rogatty: This will fail for a coupon with 0% but I guess it's fine 🤷‍♂️

resources/views/payment/cart.blade.php
```diff
@@ -0,0 +1,33 @@
+
+<div class="card column cart{{$hasCoupon ? ' has-coupon' : ''}}">
+    @if(!empty($productName))
```
rogatty: This is needed only for the `select-product` route, yes? If so, I'd remove this check, no need to support an obsolete case and increasing complexity.

resources/views/payment/cart.blade.php
```diff
@@ -0,0 +1,33 @@
...
+            <span>@lang('payment.cart-shipment-value')</span>
+        </p>
+        @if(!empty($hasCoupon))
+            <p>
+                <span>@lang('payment.cart-coupon-label')</span><span>{{$couponValue}}</span>
+            </p>
+        @endif
+        @if (empty($hasCoupon))
+            <span>@lang('payment.cart-price-label')</span><span>@lang('payment.cart-price-value', ['value' => $productPrice])</span>
+        @else
```
kvas-damian: you can combine it with the `if` above.

resources/views/payment/personal-data.blade.php
```diff
@@ -2,6 +2,15 @@
 
 @section('content')
 
+	@include('payment.cart', [
+		'productName' => $product->name,
+		'productPrice' => $product->price,
+		'productAccessEnd' => $product->access_end,
```
kvas-damian: Maybe we could pass entire `product` object instead of separate attributes?

resources/views/payment/confirm-order.blade.php
kvas-damian: This value can be calculated on based on `couponValue`. Do we need it?

gitkuba: `(bool) $order->coupon`

app/Http/Controllers/Payment/ConfirmOrderController.php
```diff
@@ -28,13 +28,27 @@ public function index(Payment $payment)
...
+		$couponValue = null;
+
+		if (!empty($coupon)) {
+			if ($coupon->is_percentage) {
```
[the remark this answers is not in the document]
bognix: I don't know, I don't want to hardcode currency in model.

resources/views/payment/cart.blade.php
bognix: I think we can start working on semantics once we start styling. The markup may change drastically then

resources/views/payment/confirm-order.blade.php
bognix: previously the condition was different for the personal-data view and confirm-order view. Now it's the same so we can move it to blade.

resources/views/payment/personal-data.blade.php
bognix: I wanted to be explicit. I can combine if you have strong arguments.

app/Http/Controllers/Payment/PersonalDataController.php
```diff
@@ -55,9 +55,26 @@ public function index(FormBuilder $formBuilder, $productSlug = null)
 
 		session()->flash('url.intended', route('payment-personal-data'));
 
+		$coupon = $this->readCoupon(Auth::user());
+		$productPriceWithCoupon = null;
+		$couponValue = null;
+
+		if (!empty($coupon)) {
+			if ($coupon->is_percentage) {
+				$value = number_format($coupon->value * $product->price / 100, 2, '.', '');
```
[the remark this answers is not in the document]
bognix: maybe by the product because the result is the final product price.

app/Http/Controllers/Payment/PersonalDataController.php
bognix: I'll resign from using number_format. I copied this logic from some old code but this introduces inconsistency in how we display things for percentage coupons and amount coupons.

resources/views/payment/cart.blade.php
bognix: Let's remove once we remove the select-product. I didn't want to make assumptions here.

app/Http/Controllers/Payment/PersonalDataController.php
```diff
@@ -146,6 +160,15 @@ protected function generateStudyBuddy($order)
 		);
 	}
 
+	protected function readCoupon($user) {
+		$userCoupon = $user ? $user->coupons->first() : null;
```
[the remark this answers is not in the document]
bognix: It throws for anon users: `Trying to get property 'coupons' of non-object`

app/Http/Controllers/Payment/ConfirmOrderController.php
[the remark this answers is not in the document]
rogatty: I don't like putting formatter into the model, a helper sounds more fitting.

resources/views/payment/confirm-order.blade.php
bognix: updated - I pass the coupon instance.

bognix: Ok, I convinced myself to pass the coupon instance but with product even though I would pass the product instance I would have to pass the `productPriceWithCoupon` either way because it's calculation is different on both views.

resources/views/payment/personal-data.blade.php
bognix: Ok, I convinced myself to pass the coupon instance but with product even though I would pass the product instance I would have to pass the `productPriceWithCoupon` either way because it's calculation is different on both views.

app/Http/Controllers/Payment/ConfirmOrderController.php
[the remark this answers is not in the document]
bognix: It's in view, we already had translations for this and I think it looks ok.

app/Http/Controllers/Payment/PersonalDataController.php
```diff
@@ -146,6 +148,15 @@ protected function generateStudyBuddy($order)
...
+		if (session()->has('coupon')) {
+			return session()->get('coupon')->fresh();
+		} else if ($userCoupon) {
```
rogatty: Let's keep the return type consistent, this could be simple `} else {`. I'm surprised that linter didn't catch it.

app/Models/Product.php
```diff
@@ -75,6 +75,19 @@ public function scopeSlug($query, $slug)
 			->first();
 	}
 
+	public function getPriceWithCoupon($coupon) {
```
rogatty:
```suggestion
	public function getPriceWithCoupon(Coupon $coupon = null) {
```

bognix: no need to import, the namespace is right :)

PR #1344, opened 2019-03-11
[PLAT-1065] Remove 'select product' step from checkout

gitkuba:
TODO:

- [x] Add album price to link in 'my orders'
- [ ] Update browser tests


app/Http/Controllers/Payment/PersonalDataController.php
```diff
@@ -21,22 +21,21 @@ class PersonalDataController extends Controller
...
+		} else {
+			$product = Product::slug($request->route('productSlug') ?? 'wnl-online');
+			Session::put('product', $product);
 		}
 
-		$product = Session::get('product');
-		if (!$product instanceof Product || !$product->available
-			|| $product->signups_close->isPast() || $product->signups_start->isFuture()) {
-			return redirect()->route('payment-select-product');
+		if (!$product instanceof Product ||
```
bognix: can a product be an instance of something else than product (except being empty)

package.json
```diff
@@ -2,7 +2,7 @@
   "private": true,
   "scripts": {
     "dev": "cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
-    "hot": "cross-env NODE_ENV=development node_modules/webpack-dev-server/bin/webpack-dev-server.js --progress --inline --hot --config=node_modules/laravel-mix/setup/webpack.config.js",
+    "hot": "cross-env NODE_ENV=development node_modules/webpack-dev-server/bin/webpack-dev-server.js --progress --inline --hot --config=node_modules/laravel-mix/setup/webpack.config.js --disableHostCheck=true",
```
bognix: could you share some context regarding this change?

resources/assets/js/components/user/MyOrders.vue
```diff
@@ -69,7 +77,22 @@ export default {
...
+			const prolongationOrders = this.orders.filter(order =>
+				order.product.slug === 'wnl-online' &&
+				order.coupon.value === 50 &&
+				order.coupon.type === 'percentage' &&
+				!order.canceled
+			);
+
+			const albumOrders = this.orders.filter(order => order.product.slug === 'wnl-album');
+
+			return albumOrders.length === 0 && prolongationOrders.length > 0;
```
bognix: we won't show the link for users who meet following criteria:
- participated in edition 1
- prolonged to edition 2 and bought album
- prolonged to edition 3

resources/assets/js/components/user/MyOrders.vue
```diff
@@ -6,6 +6,13 @@
 					Twoje zamówienia
 				</div>
 			</div>
+			<div v-if="displayAlbumLink && hasOrders">
```
bognix: do you need to check for `hasOrders` as well? In displayAlbumLink you already check if user has an order.

resources/assets/js/components/user/MyOrders.vue
```diff
@@ -69,7 +77,22 @@ export default {
 		},
 		userFriendlySubscriptionDate() {
 			return moment(this.currentUserSubscriptionDates.max*1000).locale('pl').format('LL');
-		}
+		},
+		displayAlbumLink() {
+			const prolongationOrders = this.orders.filter(order =>
+				order.product.slug === 'wnl-online' &&
```
bognix: maybe it's not for now but I think it would be better to check if user bought one of previous products. Checking if order has a 50% discount is a side effect of buying previous product.

resources/views/payment/signups-closed.blade.php
```diff
@@ -0,0 +1,51 @@
...
+					Brak miejsc :(<br> <a href="https://wiecejnizlek.pl/zostaw-e-mail">Kliknij i zostaw swój e-mail</a>, aby zarezerwować miejsce na kolejnej edycji!
+				</div>
+			</div>
+		@endif
+	</div>
+</section>
+@endsection
+
+@section('payment-scripts')
+	<script>typeof fbq === 'function' && fbq('track', 'AddToCart')</script>
```
bognix: are you sure we want to track "addToCart" here - user didn't add anything to cart on this page.

package.json
gitkuba: https://github.com/webpack/webpack-dev-server/issues/1604

app/Http/Controllers/Api/PrivateApi/ProductsApiController.php
```diff
@@ -36,6 +36,17 @@ public function post(CreateProduct $request)
 		return $this->transformAndRespond($product);
 	}
 
+	public function getBySlug($slug)
```
bognix: I wonder if Laravel dependency injection would work here since we have a slug defined as a scope on model.

package.json
gitkuba: Updating weback is probably a better solution, as it seems, that problem was fixed. I'll check.

app/Http/Controllers/Payment/PersonalDataController.php
gitkuba: collection? string? array? who knows? I think my uncertainty comese from the fact, that we fetch & unserialize it from session.

bognix: I wonder if we shouldn't assume that if something ducks like a ducks it is a duck and if for some reason something wrong will be stored in the session under product key we can fix the place that raised this issue.

rogatty: It's a bit defensive but IMO it makes sense to validate input data, even if the input is the session storage.

routes/web.php
```diff
@@ -16,7 +16,8 @@
 Auth::routes();
 
 Route::group(['namespace' => 'Payment', 'prefix' => 'payment', /*'middleware' => 'payment'*/], function () {
-	Route::get('select-product', 'SelectProductController@index')->name('payment-select-product');
+	Route::redirect('/select-product', '/payment/personal-data', 302)->name('payment-select-product');
+	Route::redirect('/', '/payment/personal-data', 302);
```
rogatty: 302 is a temporary redirect, I think it should be 301 in both places.

app/Models/Product.php
```diff
@@ -78,7 +78,6 @@ public function scopeSlug($query, $slug)
 	public function getAvailableAttribute()
 	{
 		return
-			$this->quantity > 0 &&
-			$this->signups_start->isPast();
+			$this->quantity > 0;
```
rogatty: How about moving these from `PersonalDataController.php` here?

```
$product->signups_close->isPast() || $product->signups_start->isFuture()
```

package.json
rogatty: I think webpack version is coupled with laravel-mix.

resources/assets/js/components/user/MyOrders.vue
```diff
@@ -69,7 +77,22 @@ export default {
...
+				order.coupon.value === 50 &&
+				order.coupon.type === 'percentage' &&
+				!order.canceled
```
rogatty: Shouldn't it be paid?

resources/assets/js/components/user/MyOrders.vue
rogatty: I think these checks should be made server-side and reused wherever we check if user can actually buy the album.

rogatty: I think these checks should be made server-side and reused wherever we check if user can actually buy the album.

resources/views/payment/signups-closed.blade.php
```diff
@@ -0,0 +1,51 @@
...
+	'title' => trans('payment.select-product-title'),
+	'subtitle' => trans('payment.select-product-subtitle'),
+])
+
+<section class="section select-product">
+	<div class="container">
+		@if(!$product)
+			<div class="column">
+				<div class="notification has-text-centered strong">
+					Zapisy zostały zakończone. <a href="https://wiecejnizlek.pl/zostaw-e-mail">Kliknij i zostaw swój e-mail</a>, aby zarezerwować miejsce na kolejnej edycji!
```
rogatty: This URL doesn't work, I think it's https://wiecejnizlek.pl/rezerwacja/ now.

app/Models/Product.php
rogatty: Ah, we can't because of https://github.com/bethinkpl/wnl-platform/pull/1344/files#diff-8c21846dc0c0655cc9810bed9a72f4b7R38

resources/assets/js/components/user/MyOrders.vue
gitkuba: Yup, that's a feature 😁

package.json
gitkuba: Current version of `webpack-dev-server` used by `laravel-mix` is `3.1.14`, which theoretically  contains a fix for that issue, but it doesn't seem to work for me.

resources/assets/js/components/user/MyOrders.vue
gitkuba: Yeah, maybe we could use explicit coupon types to recognize prolongation orders.

app/Http/Controllers/Api/PrivateApi/ProductsApiController.php
gitkuba: I'm not sure what's on your mind 🙂

resources/assets/js/components/user/MyOrders.vue
gitkuba: Doesn't have to be 🙂

resources/assets/js/components/user/MyOrders.vue
gitkuba: I'd say these are conditions we've chosen to use in this context, but it's not a rule of thumb for albums 🙂

resources/assets/js/components/user/MyOrders.vue
bognix: I didn't even mean to check for the coupon presence. I meant to check if there is an order for a product that was an online or onsite course in the past.

resources/assets/js/components/user/MyOrders.vue
bognix: Based on my understanding user can buy product only once and this is the condition we want to use everywhere. I assume that based on the copy provided by Adam recently: 

> Przedłużenie dostępu z 50% zniżką nie obejmuje nowego albumu map myśli. Nowy album możesz jednokrotnie zamówić po zalogowaniu na platformę, w zakładce KONTO > Twoje zamówienia.

resources/assets/js/components/user/MyOrders.vue
bognix: that's what I thought. Introducing product type may be useful one day.

PR #1345, opened 2019-03-11
 PLAT-641 | Stop using hardcoded payment schedule

rogatty:
https://bethink.atlassian.net/browse/PLAT-641

TODO:

- [x] Add tests


resources/assets/js/payment.js
```diff
@@ -57,6 +58,11 @@ $(function () {
 		});
 	});
 
+	$('[data-instalment-due-date]').each(function () {
+		let date = moment.unix($(this).attr('data-timestamp'));
+		$(this).html(date.format('D MMMM'));
```
rogatty: I wanted to format it server-side but Carbon requires `locales` to be installed (https://stackoverflow.com/a/32594258) and it's not a part of APK index that alpine uses and I couldn't rebuild the docker image:
```
Step 3/4 : RUN apk --no-cache add freetype-dev libjpeg-turbo-dev libpng-dev libzip-dev locales
 ---> Running in 05f17ea06d3d
fetch http://dl-cdn.alpinelinux.org/alpine/v3.9/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.9/community/x86_64/APKINDEX.tar.gz
ERROR: unsatisfiable constraints:
  locales (missing):
    required by: world[locales]
ERROR: Service 'php' failed to build: The command '/bin/sh -c apk --no-cache add freetype-dev libjpeg-turbo-dev libpng-dev libzip-dev locales' returned a non-zero code: 1
```

app/Models/Order.php
```diff
@@ -113,14 +113,19 @@ public function getInstalmentsAttribute()
...
-		// TODO: https://bethink.atlassian.net/browse/PLAT-641
-		$paymentDates = [
-			$this->created_at->addDays(7),
-			Carbon::createFromDate(2018, 11, 20),
-			Carbon::createFromDate(2018, 12, 20),
-		];
+		$orderInstalments = $this->orderInstalments()->get();
+
+		if ($orderInstalments->count() > 0) {
+			$orderInstalments = $this->generatePaymentSchedule();
```
kvas-damian: Why do we need regenerate instalments when they're already defined in DB?

resources/assets/js/payment.js
kvas-damian: you could try the solution provided in https://github.com/gliderlabs/docker-alpine/issues/144

app/Models/Order.php
gitkuba: Yup, this line seems to be redundant.

app/Models/Order.php
```diff
@@ -162,37 +167,53 @@ public function getInstalmentsAttribute()
 		];
 	}
 
+	/**
```
gitkuba: The above calculation should be entirely replaceable by the information from DB now.

gitkuba: Uhm, sorry. Couldn't get a right preview. I mean those lines: https://github.com/bethinkpl/wnl-platform/blob/7742092fc3f38890e8250add04a776653217fae9/app/Models/Order.php#L129-L160

app/Models/Order.php
```diff
@@ -113,14 +113,19 @@ public function getInstalmentsAttribute()
...
 		$nextPayment = null;
-		$now = Carbon::now();
 
-		// TODO: https://bethink.atlassian.net/browse/PLAT-641
-		$paymentDates = [
-			$this->created_at->addDays(7),
-			Carbon::createFromDate(2018, 11, 20),
-			Carbon::createFromDate(2018, 12, 20),
-		];
+		$orderInstalments = $this->orderInstalments()->get();
```
bognix: why do you call `get` here, can't you just use `$this->orderInstalments`?

app/Models/Order.php
```diff
@@ -162,37 +167,53 @@ public function getInstalmentsAttribute()
...
-			$valueToDistribute -= $amount;
-
-			if ($paidToDistribute >= $amount) {
-				$paidAmount = $amount;
-				$paidToDistribute -= $amount;
-			} else {
-				$paidAmount = $paidToDistribute;
-				$paidToDistribute = 0;
-			}
+		return $this->product->instalments()
```
bognix: how about adding unit tests to this function?

resources/assets/js/payment.js
rogatty: It's messy and I don't think is worth it. The current solution works just fine as these dates are hidden in a collapsed container anyway. I'd circle back to the topic if we ever want to optimize by removing moment.

app/Models/Order.php
rogatty: No, because `$this->generatePaymentSchedule()` returns a `Collection` object and I want to keep the interface consistent.

bognix: what is returned then by `$this->orderInstalments` ?

app/Models/Order.php
```diff
@@ -109,90 +110,79 @@ public function getCouponAmountAttribute()
...
-					'amount' => $instalment['left'],
-					'date'   => $date,
-				];
-			}
-			$totalLeft += $instalment['left'];
-		}
+		$orderInstalments = $this->generateAndSavePaymentSchedule();
+
+		$nextPayment = $orderInstalments
+			->sort(function (OrderInstalment $a, OrderInstalment $b) {
```
kvas-damian: Why do we need to sort them? Aren't they sorted when querying `ProductInstalments` model?

resources/views/payment/invoices/final.blade.php
```diff
@@ -149,20 +150,21 @@
...
+						<td>{{ $n($instalment->amount) }}zł</td>
 					</tr>
 				@endif
 			@endforeach
 			<tr>
 				<td><strong>Razem</strong></td>
-				<td>{{ $n($invoiceOrder->instalments['total']/ 1.23) }}zł</td>
-				<td>{{ $n($invoiceOrder->instalments['total']- $invoiceOrder->instalments['total'] / 1.23) }}zł</td>
+				<td>{{ $n($invoiceOrder->instalments['total'] / 1.23) }}zł</td>
+				<td>{{ $n($invoiceOrder->instalments['total'] - $invoiceOrder->instalments['total'] / 1.23) }}zł</td>
```
kvas-damian: For the future: we should avoid such calculations in view files.

app/Models/Order.php
```diff
@@ -109,90 +110,79 @@ public function getCouponAmountAttribute()
...
+			'instalments' => $orderInstalments,
 			'nextPayment' => $nextPayment,
 			'total'       => $totalLeft,
 		];
 	}
 
+	/**
+	 * @return OrderInstalment[]|\Illuminate\Support\Collection
+	 */
 	public function generatePaymentSchedule()
```
kvas-damian: It's a nice candidate for a test. Especially with coupons applied.

PR #1348, opened 2019-03-13
PLAT-1065, PLAT-1066, PLAT-1068

kvas-damian:
Run `php artisan migrate && php artisan data-migration:coupon-fill-kinds`


app/Models/User.php
```diff
@@ -259,9 +259,28 @@ public function getFullAddressAttribute()
...
+	}
+
+	public function getHasAccessAttribute() {
+		return !($this->first_name === null || $this->last_name === null);
+	}
+
 	protected function getSubscriptionStatus($dates)
 	{
-		if ($this->hasRole(['admin', 'moderator', 'test'])) return self::SUBSCRIPTION_STATUS_ACTIVE;
+		if ($this->hasRole([Role::ROLE_ADMIN, Role::ROLE_MODERATOR, 'test'])) {
```
gitkuba: `'test'` -> `Role::ROLE_TEST` ?

database/migrations/2019_03_12_104810_user_name_nullable.php
```diff
@@ -0,0 +1,34 @@
...
+use Illuminate\Database\Migrations\Migration;
+
+class UserNameNullable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
```
gitkuba: Indentation seems to be messed up 🙂

resources/views/payment/account-continue.blade.php
```diff
@@ -0,0 +1,23 @@
+@extends('payment.layout')
+
+@section('content')
+	@include('payment.payment-hero', [
+		'step' => 1,
+		'title' => 'Lorem ipsum',
```
gitkuba: Did you left this on purpose?

kvas-damian: yes, the entire hero needs to be reworked so I put a crap text there for now.

app/Http/Controllers/AppController.php
```diff
@@ -2,12 +2,19 @@
...
 
+use Illuminate\Support\Facades\Auth;
 use JavaScript;
 
 class AppController extends Controller
 {
 	public function index()
 	{
+		$user = Auth::user();
+		if (!$user->signUpComplete) {
```
rogatty: Can `$user` be null in an Http Controller context?

app/Http/Controllers/Payment/AccountController.php
```diff
@@ -0,0 +1,70 @@
...
+		Log::notice('Creating user account');
+		$user = User::create(
+			[
+				'email'              => $request->get('email'),
+				'password'           => bcrypt($request->get('password')),
+				'consent_terms'      => 1,
+			]
+		);
+
+		Mail::to($user)->send(new SignUpConfirmation($user));
```
rogatty: Shouldn't this be async?

rogatty: Ah, Laravel makes it async 👍

app/Http/Controllers/Payment/AccountController.php
```diff
@@ -0,0 +1,70 @@
...
+			if ($user->signUpComplete) {
+				return view('payment.account-name', [
+					'user' => $user,
+				]);
+			} else {
+				return view('payment.account-continue');
+			}
+		}
+
+		$request->session()->flash('url.intended', route('payment-personal-data'));
```
rogatty: I think we can either remove it or change https://github.com/bethinkpl/wnl-platform/pull/1348/files#diff-cfc0b3a92c36b8de61b2f525d21a2d88R68 to use `redirectToIntended()`

app/Http/Controllers/AppController.php
gitkuba: If user is not logged in, yes, but this controller is behind `auth` middleware, so it shouldn't be possible.

app/Http/Controllers/Payment/PersonalDataController.php
```diff
@@ -64,59 +48,58 @@ public function index(FormBuilder $formBuilder, $productSlug = null)
...
 
 		if (!$form->isValid()) {
 			Log::notice('Sing up form invalid, redirecting...');
 
 			return redirect()->back()->withErrors($form->getErrors())->withInput();
 		}
 
-		if ($user) {
-			$this->updateAccount($user, $request);
+		$user = Auth::user();
```
rogatty: This is redundant, we load user a few lines above.

gitkuba: fixed

app/Http/Controllers/Payment/PersonalDataController.php
```diff
@@ -157,62 +140,47 @@ protected function readCoupon($user) {
...
+			$userData['last_name'] = $request->last_name ?? $user->last_name;
+		}
+
+		$user->update($userData);
+
+		$user->profile()->updateOrCreate(
+			['user_id' => $user->id],
+			[
+				'first_name' => $user->first_name,
+				'last_name'  => $user->last_name,
```
rogatty: No need to update it if the form fields are disabled, right?

app/Http/Forms/PersonalDataForm.php
```diff
@@ -0,0 +1,168 @@
...
+		return parent::validate($validationRules, $messages);
+	}
+
+	protected function getIdentityNumberValidator($identityNumberType) {
+		$validators = [
+			'passport_number' => new ValidatePassportNumber,
+			'personal_identity_number' => new ValidatePersonalIdentityNumber,
+		];
+
+		if (!array_key_exists($identityNumberType, $validators)) {
```
rogatty: We have rule `required|in:personal_identity_number,passport_number` so this should never happen. But if it does, I'd prefer to fail loudly than allow users to save incorrect data.

rogatty:
```suggestion
		// If someone sends invalid `identity_number_type` then disable `identity_number` validator
		// Request will fail anyway
		if (!array_key_exists($identityNumberType, $validators)) {
```

app/Models/User.php
```diff
@@ -352,6 +371,27 @@ public function sendPasswordResetNotification($token)
 		$this->notify(new ResetPasswordNotification($token));
 	}
 
+	public static function createWithProfileAndBilling($userData) {
+		/** @var User $user */
+		$user = static::create($userData);
+
+		$user->profile()->firstOrCreate([
```
rogatty: `firstOrCreate` will search DB for profile with the given attributes, I don't see a reason to do that. We can use `create()` here and for the `billing` below.

database/migrations/2019_03_12_104810_user_name_nullable.php
```diff
@@ -0,0 +1,34 @@
...
+
+	/**
+	 * Reverse the migrations.
+	 *
+	 * @return void
+	 */
+	public function down()
+	{
+		Schema::table('users', function (Blueprint $table) {
+			$table->longText('first_name')->change();
```
rogatty: This will fail if there was a new user created after the migration. Might be better to just do nothing on `down()`

resources/views/app_unavailable.blade.php
```diff
@@ -34,7 +34,7 @@
 			</span>
 
 			<div class="nav-right nav-menu">
-				<a href="https://platforma.wiecejnizlek.pl/payment/select-product" class="nav-item">
+				<a href="https://platforma.wiecejnizlek.pl/payment/account" class="nav-item">
```
rogatty: Let's not hardcode `platforma`

resources/views/mail/sign-up-confirmation.blade.php
```diff
@@ -10,17 +10,7 @@
 	<p style="font-size: 1.1rem; font-weight: bold;">Przydatne linki</p>
```
rogatty: There is only one link.

resources/views/mail/sign-up-confirmation.blade.php
```diff
@@ -10,17 +10,7 @@
...
-			</a>
-		</li>
-		<li>
-			<a href="https://platforma.wiecejnizlek.pl/app/myself/orders">
-				Konto > Twoje zamówienia
-			</a>
-		</li>
-		<li>
-			<a href="https://platforma.wiecejnizlek.pl/payment/select-product">
+			<a href="{{url('payment/account')}}">
```
rogatty: It's weird that we send users to payment. What if they already finished the order and then checked the email (common use case, I'd say)? Let's consult with Luke.

resources/views/app_unavailable.blade.php
gitkuba: This file is used on demo, hence generating url is not an option 🤷‍♂️

app/Models/User.php
gitkuba: fixed

database/migrations/2019_03_12_104810_user_name_nullable.php
gitkuba: fixed.

resources/views/mail/sign-up-confirmation.blade.php
gitkuba: done.

resources/views/mail/sign-up-confirmation.blade.php
gitkuba: I think it's ok, but requires a better description.

resources/views/payment/confirm-order.blade.php
```diff
@@ -40,31 +40,41 @@
...
 
-		<section class="subsection">
-			<div class="box">
-				<p class="title">@lang('payment.confirm-personal-data-heading')</p>
-				<p class="big">{{ $user->full_name }}</p>
-				<p><strong>{{ $user->email }}</strong></p>
-
-				<p class="margin top big">@lang('payment.confirm-personal-data-private')</p>
+		@if($order->has_shipment)
+			@if ($order->has_shipment)
```
rogatty: This check is redundant, user will never see `@lang('payment.confirm-no-album-info')`

resources/views/payment/confirm-order.blade.php
```diff
@@ -40,31 +40,41 @@
...
+								<strong>
+									{{ trans('payment.identity_number_' . $user->identityNumbers[0]['type']) }}:
+									{{ $user->identityNumbers[0]['value'] }}
+								</strong>
+							</p>
+						@endif
 
-				@if(is_array($user->identityNumbers))
+						<p class="margin top big">@lang('payment.confirm-personal-data-address')</p>
+					<ul>
```
rogatty: Indentation is wrong.

resources/views/payment/signups-closed.blade.php
```diff
@@ -0,0 +1,47 @@
...
+			<div class="column">
+				<div class="notification has-text-centered strong">
+					Zapisy zostały zakończone. <a href="https://wiecejnizlek.pl/rezerwacja/">Kliknij i zostaw swój e-mail</a>, aby zarezerwować miejsce na kolejnej edycji!
+				</div>
+			</div>
+		@endif
+		@if($product->signups_close->isPast())
+			<div class="column">
+				<div class="notification has-text-centered strong">
+					Zapisy zostały zakończone. <a href="https://wiecejnizlek.pl/rezerwacja/">Kliknij i zostaw swój e-mail</a>, aby zarezerwować miejsce na kolejnej edycji!
```
rogatty: This text is used twice here and above. Is this correct? If so, we can use single `if` block.

tests/Browser/Tests/Payment/Modules/AccountModule.php
```diff
@@ -0,0 +1,64 @@
...
+	{
+		$this->navigate($browser);
+
+		$browser->on(new AccountPage());
+
+		$accountData = $this->generateAccountFormData();
+		$this->fillInAccountForm($accountData, $browser);
+		$browser->accountData = $accountData;
+
+		$browser->xpathClick('.//button[@class="button is-primary"]');
```
rogatty: For instalments tests I've added `data-button` attribute to the buttons. I think it's more future proof, this will probably fail as soon as we apply new styles.

tests/Browser/Tests/Payment/Modules/AccountModule.php
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	public function logInUsingModal(BethinkBrowser $browser)
+	{
+		if (!empty($browser->user)) return __CLASS__;
+
+		$faker = new Generator();
+		$faker->addProvider(new Person($faker));
+		$user = factory(\App\Models\User::class)->create();
+		$browser->accountData = $user;
```
rogatty: This is weird, sometimes we store array with form data and sometimes `User` instance. Can we keep it consistent?

tests/Browser/Tests/Payment/Modules/AccountModule.php
```diff
@@ -0,0 +1,64 @@
...
+			->visit('/')
+			->assertPathIs(
+				(new AccountPage)->url()
+			);
+	}
+
+	protected function navigate(BethinkBrowser $browser)
+	{
+		// Check if personal-data is current page, if not,
+		// navigate to it using a product.
```
rogatty: This comment is outdated.

PR #1350, opened 2019-03-13
Update content of e-mails and introduce a new one for Study Buddy

adamkarminski:
This PR updates some misleading content and expired links. I didn't have time to test the content though, so please check the changed e-mails and make changes if necessary.

I created another ticket for sending an extra e-mail when somebody gets a study buddy coupon: https://bethink.atlassian.net/browse/PLAT-1119

### DIFF
--- a/resources/views/mail/sign-up-confirmation.blade.php
+++ b/resources/views/mail/sign-up-confirmation.blade.php
@@ -7,16 +7,21 @@
 
 	<p>Jeśli wszystko poszło po naszej myśli, mail potwierdzający zamówienie zaraz pojawi się w Twojej skrzynce! :)</p>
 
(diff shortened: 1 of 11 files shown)


resources/views/mail/sign-up-confirmation.blade.php
```diff
@@ -7,23 +7,18 @@
...
 			</a>
 		</li>
 		<li>
-			<a href="https://platforma.wiecejnizlek.pl/app/myself/orders">
+			<a href="{{url('app/myself/orders')}}">
 				Konto > Twoje zamówienia
 			</a>
 		</li>
-		<li>
-			<a href="https://platforma.wiecejnizlek.pl/payment/select-product">
```
kvas-damian: As we discussed offline, we're are going to send this email immediately after account creation. Such user can't access platform and orders. We should bring back him to the checkout process.

PR #1353, opened 2019-03-16
Improve build performance

gitkuba:
**Optimization no. 0**: Adjust the number of CPUs

| vCPUs | php install           | js install  | js build | rest* | total
| ------------- |-------------| -----| --- | --- | --- |
| 1      | 2:32 | 2:01 | 3:42 | 3:52 | **12:08**
| 2     | 1:55 |   1:01 | 2:48 | 1:35 | **7:31**
| 4     | 2:00 |    1:00| 2:03 | 1:34 | **6:39**

Paying for 4CPUs doesn't seem to be worth it, so I left it with 2.

**Optimization no. 1**: Make better use of docker layers to cache php dependencies + use composer plugin to parallelize package downloads = 2 minutes saved 🎉

| vCPUs | php install  | js install  | js build | rest* | total
| --- | --- | --- | --- | --- | --- |
| 2   |  0:04 | 0:51 | 3:26 | 1:36 | **5:58**

**Optimization no. 2**: Cache `node_modules` = 1 minute saved 🎉

| vCPUs | php install  | js install  | js build | rest* | total
| --- | --- | --- | --- | --- | --- |
| 2   |  0:04 | 0:26 | 2:55 | 1:21 | **4:36**

Seems to work well also via ansible - the whole playbook takes ca. 6min to run vs 9min w/o optimizations.

TODO: 
- [x] Time table before dockerfile optimization
- [x] Check if we can cache compiled js
- [x] Check why build via ansible still takes ca. 8 min (as if no layer cache was used)

Test cases:
- [x] Try to update php dependencies
- [x] Try to update js dependencies
- [x] try to update app code (vue & php)
- [x] check the critical app features

Inspiration:
https://medium.com/@c.harrison/speedy-composer-installs-in-docker-builds-41eea6d0172b
https://nickjanetakis.com/blog/docker-tip-2-the-difference-between-copy-and-add-in-a-dockerile

*newrelic, mv, chmod...


gitkuba:
@bethinkpl/platform_devs Ok, I'm done for now. Everything seems to work fine. There's certainly a field for further improvements:
- js minifier plugin (takes most of the time) is capable of using only 2 CPUs, maybe it's doable to make it use more resources. Also a VM with stronger CPU would do the job (maybe running it on vlad isn't a bad idea after all; is there a way to compile JS using GPU?) 
- chmod/rm/cp operations seem to be suspiciously time consuming
- we currently push to 2 docker registries (gcr and bethink)

PR #1357, opened 2019-03-22
[PLAT-1135] Make sure coupon is being updated

gitkuba:


app/Http/Controllers/Payment/PersonalDataController.php
```diff
@@ -49,7 +49,7 @@ public function handle(Request $request)
 		$this->updateAccount($user, $request, $form);
 
 		if (!!Session::get('orderId')) {
-			$this->updateOrder($product, $user, $request);
+			$this->updateOrder($product, $user, $request, $coupon);
 		} else {
 			$this->createOrder($product, $user, $request);
```
rogatty: We could pass the coupon here too. No need to read it again inside.

PR #1360, opened 2019-03-22
[PLAT-1142] Fix checkout crashing for prolonging user

gitkuba:
TODO:
- [x] Fix browser tests


resources/views/payment/personal-data.blade.php
```diff
@@ -53,7 +53,7 @@
 		</section>
 
 		<section class="o-checkoutSection">
-			@if(empty($coupon) || $coupon->kind !== \App\Models\Coupon::KIND_PARTICIPANT)
+			@if(empty($coupon) || ($coupon && $coupon->kind !== \App\Models\Coupon::KIND_PARTICIPANT))
```
rogatty: This change isn't needed, PHP won't execute the second condition if `empty($coupon) === true`.

tests/Browser/Tests/Payment/Modules/MyOrdersModule.php
```diff
@@ -35,7 +35,9 @@ public function assertOrderPlaced(BethinkBrowser $browser)
 			->waitForText('Twoje zamówienia', 60)
 			->waitForText('Zamówienie numer ' . $order->id);
 
-		if (empty($browser->coupon)) {
+		if (empty($browser->coupon) ||
+			(!empty($browser->coupon) && $browser->coupon->kind !== App\Models\Coupon::KIND_PARTICIPANT)
```
rogatty:
```
if (empty($browser->coupon) || $browser->coupon->kind !== App\Models\Coupon::KIND_PARTICIPANT)
```
would be enough

PR #1362, opened 2019-03-26
Install Laravel Telescope

gitkuba:


rogatty:
Tests failed with:
```
[2019-04-01 08:20:18] testing.ERROR: No matching handler found for Mockery_1_Illuminate_Redis_RedisManager::set('UserState:Course:1:1:1', '"foo bar"'). Either the method was unexpected or its arguments matched no expected argument list for this method
 {"userId":1,"email":"kuba.karminski@bethink.pl","exception":"[object] (Mockery\\Exception\\NoMatchingExpectationException(code: 0): No matching handler found for Mockery_1_Illuminate_Redis_RedisManager::set('UserState:Course:1:1:1', '\"foo bar\"'). Either the method was unexpected or its arguments matched no expected argument list for this method
```
This is something I saw previously and it's caused by upgrading Mockery. I suggest bumping it back down and creating a ticket to unblock the upgrade.

gitkuba:
I'm not really convinced to this package after testing its abilities. Mostly I was keen to see queue monitoring feature, but it seems to be very basic and not that helpful.

PR #1361, opened 2019-03-22
PLAT-998 | Onboarding

rogatty:
https://bethink.atlassian.net/browse/PLAT-998

TODO:
- [x] save the last step user was on to DB
- [x] block access to other routes if user didn't finish onboarding; redirect to their last onboarding step
- [x] limit items shown in the main navigation
- [x] add "next step" buttons
- [x] add stepper
- [x] handle navigation by browser back button
- [x] validate that user went through all steps; block jumping more than one step
- [x] when user finishes the last step, mark onboarding as finished
- [x] `/app/onboarding` should redirect to the user's last step
- [x] render user plan component
- [x] disable submit button on user plan after saving; re-enable on change
- [x] style all the steps
- [x] show different text for returning users on the user-plan step
- [x] display onboarding only when subscription "status active"
- [x] users using previous edition shouldn't see the onboarding
- [x] add loading state for buttons
- [x] aggregate four blocking requests into one
- [x] add tests
- [x] cleanup `Onboarding.vue` and remove hacks
- [x] update copies after Ola's review
- [ ] if https://github.com/bethinkpl/wnl-platform/pull/1365 is merged first, check if we won't break anything from there
- [ ] if https://github.com/bethinkpl/wnl-platform/pull/1363 is merged first, check if we won't break anything from there
- [x] ~add dashboard news~
- [x] ~button linking to "Wstępny LEK" should also work for LDEK~ https://bethink.atlassian.net/browse/PLAT-1152
- [x] ~hide the user plan step for LDEK~ https://bethink.atlassian.net/browse/PLAT-1152

PR #1363, opened 2019-03-26
[PLAT-1148] Provide immediate access to platform

gitkuba:


app/Jobs/CreateSubscription.php
```diff
@@ -0,0 +1,58 @@
...
+use App\Models\Order;
+use App\Models\User;
+use App\Models\UserSubscription;
+use Illuminate\Queue\SerializesModels;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+
+class CreateSubscription
+{
+    use Dispatchable, InteractsWithQueue, SerializesModels;
```
bognix: fix indent.

bognix:
@gitkuba this doesn't fix the initial problem so if the payment is processed we don't refresh the subscription data on the client side.

app/Jobs/CreateSubscription.php
```diff
@@ -0,0 +1,58 @@
...
+			? min([$subscriptionAccessStart, $product->access_start])
+			: $product->access_start;
+		$accessEnd = max([$subscriptionAccessEnd, $product->access_end]);
+
+		UserSubscription::updateOrCreate(
+			['user_id' => $user->id],
+			['access_start' => $accessStart, 'access_end' => $accessEnd]
+		);
+
+		\Cache::forget(User::getSubscriptionKey($this->order->user->id));
```
rogatty: If https://github.com/bethinkpl/wnl-platform/pull/1361 gets merged first, update it to:
```suggestion
		\Cache::forget(UserSubscription::getCacheKey($this->order->user->id));
```

app/Observers/OrderObserver.php
```diff
@@ -32,6 +33,7 @@ public function updated(Order $order)
 				$order->paid = true;
 				$order->paid_at = Carbon::now();
 				$order->save();
+				$this->dispatchNow(new CreateSubscription($order));
```
rogatty: Why is it inside the `if` block and not below `$this->dispatch(new OrderPaid($order));`? The logic was different previously.

rogatty:
Following Bogna's comment - we should either reload the page in https://github.com/bethinkpl/wnl-platform/blob/a43093bfeeb23a9f8cbaf1df7e13909f50591163/resources/assets/js/components/user/Order.vue#L572-L575
 or refetch the requests that fail without active subscription:
- `/course_structure_nodes/1?include=groups,lessons,course,structurable`
- `/chat_rooms/.getPrivateRooms?include=profiles&limit=20&page=`

resources/assets/js/components/App.vue
```diff
@@ -111,14 +110,7 @@ export default {
 				});
 
 				// Setup Chat
-				const userChannel = 'authenticated-user';
-				this.fetchUserRoomsWithMessages({page: 1})
-					.then((pointer) => this.$socketJoinRoom(userChannel, pointer))
-					.then((data) => {
-						this.updateFromEventLog(data.events);
-						this.setConnectionStatus(true);
-						this.$socketRegisterListener(SOCKET_EVENT_USER_SENT_MESSAGE, this.onNewMessage);
```
rogatty: It's no longer used in this component so we can remove the import.

resources/assets/js/store/modules/chatMessages.js
```diff
@@ -288,6 +292,16 @@ const actions = {
...
 	},
+
+	async setupChat({dispatch}) {
+		dispatch('setConnectionStatus', false);
+		const userChannel = 'authenticated-user';
+		const pointer = await dispatch('fetchUserRoomsWithMessages', {page: 1});
+		const data = await this._vm.$socketJoinRoom(userChannel, pointer);
+		dispatch('updateFromEventLog', data.events);
+		dispatch('setConnectionStatus', true);
+		this._vm.$socketRegisterListener(SOCKET_EVENT_USER_SENT_MESSAGE, (msg) => dispatch('onNewMessage', msg));
```
bognix: `this._vm` looks a bit hacky. It's an internal property and shouldn't be used by the application code.

resources/assets/js/store/modules/chatMessages.js
```diff
@@ -288,6 +292,16 @@ const actions = {
 			}
 		});
 	},
+
+	async setupChat({dispatch}) {
```
bognix: I would consider moving it to plugin. Plugin is aware of the store and it can dispatch actions.

bognix: See how the plugin commits `SOCKET_CONNECTION_ERROR` mutation and what happens next. You could do similar thing so create a method called "$socketConnectionSetup" which will take the userChannel as the argument.

Then inside this method you can either dispatch the `fetchUserRoomsWithMessages` action or dispatch some generic action like `SOCKET_CONNECTION_SETUP` which then will be handled by store and proper action would be dispatched.
Inside this setup method you then call the $socketJoinRoom which you have access to because you are already inside the plugin and then you dispatch those two methods in the same manner as the `fetchUserRoomsWithMessages`

resources/assets/js/store/modules/chatMessages.js
gitkuba: I haven't found any other working solution allowing me to access call plugin method from store. @kvas-damian @rogatty any ideas?

resources/assets/js/store/modules/chatMessages.js
gitkuba: As you mentioned, chat plugin will no longer be application-agnostic then. Right now it's a nice socket.io abstraction and i'm not sure that putting there logic tightly bound with our private chat feature is clearly less hacky than using `this._vm`. @kvas-damian @rogatty what's your opinion?

resources/assets/js/store/modules/chatMessages.js
rogatty: If we want to release it today `this._vm` looks good enough. We could also pass it in, see https://github.com/vuejs/vuex/issues/1399

bognix: I mentioned it in other comment, it's easy to move this logic to plugin and doesn't require to use internal properties. Passing vue instance to vuex action as an argument is hacky as well.

PR #1369, opened 2019-03-28
PLAT-1122 | Don't show optional lessons in manual planner

kvas-damian:


database/migrations/2019_03_28_192929_delete_optional_user_lessons.php
```diff
@@ -0,0 +1,26 @@
...
+class DeleteOptionalUserLessons extends Migration
+{
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up()
+	{
+		DB::statement('DELETE FROM user_lesson WHERE lesson_id IN (SELECT id FROM lessons WHERE is_required = 0);');
```
rogatty: Let's confirm it with Darek if we haven't yet.

kvas-damian: Confirmed!

PR #1380, opened 2019-04-01
PLAT-1102 | Remove display name

rogatty:
https://bethink.atlassian.net/browse/PLAT-1102


kvas-damian:
After this change deleted accounts will be presented on frontend as: "account deleted". I think we should handle it to use Polish strings.

app/Models/User.php
```diff
@@ -162,7 +162,13 @@ public function userProductStates() {
 
 	public function getFullNameAttribute()
 	{
-		return $this->first_name . ' ' . $this->last_name;
+		$fullName = $this->first_name . ' ' . $this->last_name;
+
+		if ($fullName === 'account deleted') {
+			$fullName = 'Konto usunięte';
```
bognix: please use translations for this one - the API should be language agnostic.

app/Models/UserProfile.php
```diff
@@ -48,21 +47,17 @@ public function getAvatarUrlAttribute()
 
 	public function getFullNameAttribute()
 	{
-		return "$this->first_name $this->last_name";
+		$fullName = $this->first_name . ' ' . $this->last_name;
```
bognix: I don't like the duplication but on the other hand loading the `user` model just to get a full name is an overkill.

rogatty:
Release on 11.04.2019

resources/assets/js/components/chat/PrivateChat.vue
```diff
@@ -80,13 +80,13 @@ export default {
...
 	computed: {
-		...mapGetters(['isOverlayVisible', 'currentUserId', 'currentUserDisplayName']),
+		...mapGetters(['isOverlayVisible', 'currentUserId']),
 		...mapGetters('chatMessages', ['getProfileByUserId', 'profiles', 'getInterlocutor']),
 		interlocutorProfile() {
 			return this.getInterlocutor(this.room.profiles);
 		},
 		chatTitle() {
-			return this.interlocutorProfile.display_name || this.currentUserDisplayName;
+			return this.interlocutorProfile.full_name;
```
kvas-damian: Do we change this fallback on purpose? Do we know if we don't use it now?

rogatty: Good catch, I brought it back.

PR #1389, opened 2019-04-02
PLAT-1161 | Fix coupons syncing

rogatty:
https://bethink.atlassian.net/browse/PLAT-1161

Remember to remove all coupons from LDEK before releasing this and update env on lek production to re-enable syncing.


app/Events/Coupons/CouponEvent.php
```diff
@@ -2,15 +2,22 @@
...
-abstract class CouponEvent {
+abstract class CouponEvent
+{
 	public $coupon;
 
-	abstract public function shouldSync();
 	abstract public function sync();
 
-	public function issueSyncRequest($method, $coupon) {
+	public function shouldSync(Coupon $coupon)
```
kvas-damian: do we need a separate method for that?

kvas-damian:
Tests are failing

app/Events/Coupons/CouponEvent.php
rogatty: We're overriding it in the child classes, so I think it makes sense.

bognix:
@rogatty how about creating another PR with migration that would mark "kind" column as not nullable in database?

PR #1396, opened 2019-04-04
[PLAT-1187] exclude users with 50% discount from shipping list

preziu:


app/Console/Commands/AddressesExport.php
```diff
@@ -59,7 +59,10 @@ public function handle()
 				->whereIn('id', $ids);
 		} elseif ($products) {
 			$products = explode(',', $products);
-			$orders = Order::where('paid', 1)
+			$orders = Order::whereDoesntHave('coupon', function ($query) {
+					$query->where('kind', 'participant');
```
gitkuba:
```suggestion
					$query->where('kind', Coupon::KIND_PARTICIPANT);
```

PR #1399, opened 2019-04-05
PLAT-1197 | disable pages which do not make sense without user access

bognix:


resources/assets/js/components/global/Page.vue
```diff
@@ -53,7 +45,13 @@ export default {
 			discussion_id: 0
 		};
 	},
+	computed: {
+		hasQna() {
+			return this.is_discussable;
```
rogatty: Let's check `discussion_id` too.

kvas-damian: Add `is_discussable` to `data`

PR #1400, opened 2019-04-05
PLAT-1192 | expose info about albums shipment delayed

bognix:


resources/assets/js/components/global/splashscreens/UpcomingEdition.vue
```diff
@@ -75,6 +79,7 @@ export default {
 		};
 	},
 	computed: {
+		...mapGetters('course', ['appInstanceName']),
```
rogatty: Why it's in the course? IMO a global is a better fit. Especially that we use one a few lines below.

bognix: because in my dreams we could return the property from the course here and I wanted to make that tiny step in this direction. We can't use global in the template, hence I would have to create a data prop on both components that use this getter, right now the access to this variable from env is only in one place.

PR #1401, opened 2019-04-05
PLAT-1143 | improve pesel validation

bognix:


app/Rules/ValidatePersonalIdentityNumber.php
```diff
@@ -27,8 +45,35 @@ public function passes($attribute, $value)
...
+			$month = $monthInCentury - 40;
+		} else if ($monthInCenturyFirstDigit === 6 || $monthInCenturyFirstDigit === 7) {
+			$century = 2200;
+			$month = $monthInCentury - 60;
+		} else if ($monthInCenturyFirstDigit === 8 || $monthInCenturyFirstDigit === 9) {
+			$century = 1800;
+			$month = $monthInCentury - 80;
+		}
+
+		$year = $century + substr($value,0,2);
```
rogatty:
```
Binary operation "+" between 1800|1900|2000|2100|2200 and string results in an error.   
```

app/Rules/ValidatePersonalIdentityNumber.php
```diff
@@ -2,18 +2,36 @@
...
 namespace App\Rules;
 
+use Carbon\Carbon;
 use Illuminate\Contracts\Validation\Rule;
 
 class ValidatePersonalIdentityNumber implements Rule
 {
+	public function message()
+	{
+		return 'PESEL jest nieprawidłowy :(';
```
rogatty: Shouldn't this be a translation?

app/Rules/ValidatePersonalIdentityNumber.php
```diff
@@ -2,18 +2,36 @@
...
 	public function passes($attribute, $value)
 	{
-		$sum = 0;
+		if (!$this->validateChecksum($value)) return false;
+
+		$birthDate = $this->getBirthDate($value);
+
+		if (!checkdate($birthDate['month'], $birthDate['day'], $birthDate['year'])) return false;
 
-		if (!preg_match('/^[0-9]{11}$/', $value)) {
```
kvas-damian: I think we're missing length validation now. I suspect it'd be possible to pass the validation if you provide correct date and checksum.

PR #1405, opened 2019-04-08
[PLAT-1113] Restore public products endpoint

gitkuba:


app/Http/Controllers/Api/PublicApi/ProductsApiController.php
```diff
@@ -0,0 +1,40 @@
...
+use Illuminate\Http\Request;
+use League\Fractal\Resource\Collection;
+
+class ProductsApiController extends ApiController
+{
+	/**
+	 * ProductsApiController constructor.
+	 * @param Request $request
+	 */
+	public function __construct(Request $request)
```
kvas-damian: Do we need this constructor? We don't use the `resourceName` that we set here.

PR #1407, opened 2019-04-08
PLAT-1165 | Improve error handling in user profile

rogatty:
https://bethink.atlassian.net/browse/PLAT-1165

```
Trying to get property 'profile' of non-object at /www/current/app/Http/Controllers/Api/PrivateApi/UserProfilesApiController.php:22
```

Fix #**1**:
We were requesting `/papi/v2/users/undefined/profile` when user opened `/app/users` - not sure where they got the link from, couldn't find anything.

Fix #**2**:
Instead of infinite loading spinner on non-existent user profile (e.g. `/app/users/2976`), users will see an error message and we'll get a Sentry log.

Fix #**3**:
`logger.capture` will actually work. It didn't log to Sentry since https://github.com/bethinkpl/wnl-platform/commit/cbfe5df63404a9fa786b1e3d14b5d16e6659ce0f because `this.useExternal()` always returned `false`:
```
var x; x < 7;
false
```
We use it in a lot of places so I suspect an influx of new Sentry items.

Fix #**4**:
Return 404 instead of 500 when user is not found.

Missing fix #**5**:
I wasn't able to find out how users end up on those broken routes with profile id in the URL instead of user id. Hopefully, the new logs will help us.


kvas-damian:
According to sentry most errors have the userID. The problem is that we don't have such users in DB.

resources/assets/js/components/users/UserProfile.vue
```diff
@@ -395,61 +405,69 @@ export default {
...
-				};
-
-				const sortedAnswersCopy = [...this.sortedAnswers];
-
-				sortedAnswersCopy.reverse().forEach((answer) => {
-					config.highlighted[answer.qna_questions] = answer.id;
-				});
-
-				this.qnaConfig = config;
+				this.allQuestions = this.loadQuestions(questionsWithIncludes);
```
kvas-damian: Don't you need to get `allQuestions.data`?

resources/assets/js/components/users/UserProfile.vue
```diff
@@ -395,61 +405,69 @@ export default {
...
+			this.setUserQnaQuestions(data);
+			const {included: _, ...allQuestions} = data;
+			return allQuestions;
 		},
-		loadQuestionsForAnswers(questionsIds) {
-			return axios.post(getApiUrl('qna_questions/byIds'), {
+		async loadQuestionsForAnswers() {
+			const questionsIds = this.sortedAnswers.map((element) => {return element.qna_questions;});
+
+			const {data} = await axios.post(getApiUrl('qna_questions/byIds'), {
```
kvas-damian: Why do we need this additional request? Can't we include QnaQuestion in the initial request?

resources/assets/js/components/users/UserProfile.vue
```diff
@@ -395,61 +405,69 @@ export default {
...
+				this.allQuestions = this.loadQuestions(questionsWithIncludes);
+				this.allQuestionsForAnswers = await this.loadQuestionsForAnswers();
+				this.qnaConfig = this.loadConfig();
 
 				this.$emit('userDataLoaded', {
 					profile: this.profile
 				});
+			} catch (exception) {
+				$wnl.logger.capture(exception);
+				this.isLoadingError = true;
```
kvas-damian: Should we reset this error in `loadData`?

resources/assets/js/components/users/UserProfile.vue
kvas-damian: Ahh I see `sortedAnswers` is limited to 2 elements. I think we should rename it.

rogatty: Agreed, renamed in https://github.com/bethinkpl/wnl-platform/pull/1407/commits/d47bf5a30c02a2b7da0390180f0b255299bd1594.

resources/assets/js/components/users/UserProfile.vue
rogatty: I don't think you can go from a broken profile to a working one without unmounting this component which resets the data. Anyway, I added it in https://github.com/bethinkpl/wnl-platform/pull/1407/commits/eef9bae92cbc0b8aa06f3fd4b5c4e0ead3213b11 to be sure.

PR #1410, opened 2019-04-08
Fix admin syncing order payment status

rogatty:
Fix https://sentry.io/organizations/wnl-production/issues/963680710/?project=1207076
```
Argument 1 passed to App\Http\Controllers\Api\Transformers\OrderTransformer::transform() must be an instance of App\Models\Order, null given
```


app/Http/Controllers/Api/PrivateApi/OrdersApiController.php
```diff
@@ -29,7 +28,12 @@ public function get($id)
 			$orders = $user->orders;
 			$resource = new Collection($orders, new OrderTransformer, $this->resourceName);
 		} else {
-			$order = $user->orders()->find($id);
+			$order = Order::find($id);
```
kvas-damian: We're not fixing the problem here.
The root cause is that admin is requesting order with a given id that doesn't belong to him. To reproduce the problem go to: https://lek.wiecejnizlek.pl/admin/app/users/2862#orders

kvas-damian: current API looks inconsistent to me.
As admin I can fetch `all` orders which are limited to me. When I provide an ID, the limitation is gone.
I think we have 2 options
a. make it clear in the API url that orders are limited to user by changing the url to `/user/current/orders/all` and revert the changes
b. return all orders for admins. 

The latter most likely will break the my-orders page for admins on platform. That's why I suggest to use path a.

app/Http/Controllers/Api/PrivateApi/UserOrdersApiController.php
```diff
@@ -106,9 +115,16 @@ public function validateCoupon(Order $order, Coupon $coupon)
 		return $errors;
 	}
 
-	public function cancel($id)
+	public function cancel($userId, $id)
 	{
-		$order = Order::find($id);
+		$user = User::fetch($userId);
+
+		if (!Auth::user()->can('view', $user)) {
```
rogatty: Do we need it if there is this check below?
```
if (!\Auth::user()->can('cancel', $order)) {
	return $this->respondForbidden();
}
```

kvas-damian: We don't but without it we need to handle not existing users.

resources/assets/js/components/user/Order.vue
```diff
@@ -501,7 +509,7 @@ export default {
 			return this.order.hasOwnProperty('studyBuddy') && this.order.studyBuddy.status !== 'expired';
 		},
 		couponUrl() {
-			return `orders/${this.order.id}/coupon`;
+			return `users/current/orders/${this.order.id}/coupon`;
```
rogatty: Shouldn't we pass user id as a prop with `current` as the default? All those URLs aren't correct when admin looks at other users' orders. I'm not sure if they're currently used but anyway they're not correct.

kvas-damian: :+1:

rogatty:
LGTM

PR #1413, opened 2019-04-09
[PLAT-922] Handle instalments for an overpaid order

gitkuba:


app/Observers/OrderObserver.php
```diff
@@ -58,6 +58,10 @@ public function updated(Order $order)
 		) {
 			$this->handleCouponChange($order);
 		}
+
+		if($order->isDirty(['paid_amount'])){
+			$order->generateAndSavePaymentSchedule();
```
rogatty: We already do this in `OrderPaid::handleInstalments`, why is it not enough? Let's add a comment explaining this.

PR #1416, opened 2019-04-10
PLAT-1218 | Add order Id to Przelewy24 description

kvas-damian:


resources/assets/js/components/user/P24Form.vue
```diff
@@ -53,6 +53,9 @@ export default {
 		},
 		amount() {
 			    return this.paymentData.amount * 100;
+		},
+		description() {
+			return `Zamówienie: ${this.order.id}, ${this.order.product.name}`;
```
rogatty: Is it something that Adam will copy and paste or maybe we should use a more parsable format like JSON?

kvas-damian: This description is visible for users.

PR #1419, opened 2019-04-10
PLAT-910 | Remove orderInstalments on paymentMethod change

kvas-damian:


resources/assets/js/components/user/Order.vue
```diff
@@ -92,6 +92,7 @@
 							</button>
 						</p>
 						<p class="metadata aligncenter margin top">Kwota do zapłaty: {{this.order.total}}zł</p>
+						<p class="aligncenter" v-if="canChangePaymentMethod">Aby zapłacić na raty, anuluj to zamówienie i złóż kolejne.</p>
```
bognix: this message in not right. You don't always want to change from "full price" to "instalments"

rogatty: We check if the method is `online` a few lines above, so it's correct.

PR #1420, opened 2019-04-11
PLAT-1105 | Send only one payment reminder + keep old reminders in DB

kvas-damian:


app/Console/Commands/OrdersHandleUnpaid.php
```diff
@@ -107,7 +103,8 @@ protected function handleUnpaidInstalment()
 			function ($query) use ($beforeDue) {
 				$query
 					->whereRaw('order_instalments.paid_amount < order_instalments.amount')
-					->whereDate('due_date', "<=", $beforeDue);
+					->whereDate('due_date', "<=", $beforeDue)
+					->whereDate('due_date', ">=", Carbon::today()->subDays(7));
```
rogatty: Why 7 days? It seems like a magic number. This command is run twice a day so it should be enough to subtract 12 hours, correct?
Is it just to be safe? If so, let's add a comment.

kvas-damian: We suspend account after 4 days from the last sent reminder, so 12h isn't enough. For safety let's take a longer period.

PR #1422, opened 2019-04-12
Fix instalments integrity check - exclude album orders

kvas-damian:


checks/DataIntegrity/InstalmentsDueDateCheck.php
```diff
@@ -8,6 +8,7 @@ public function check() {
 			->join('order_instalments', 'orders.id', '=', 'order_instalments.order_id')
 			->join('products', 'orders.product_id', '=', 'products.id')
 			->whereRaw('order_instalments.due_date < products.signups_start')
+			->whereRaw('orders.product_id <> 8')
```
rogatty: We already join with `products`, so maybe use the `wnl-album` slug?

kvas-damian: What if someone assign slug to another product? We don't accept instalments now for album so new product shouldn't be affected.

rogatty: What about LDEK?

checks/DataIntegrity/InstalmentsDueDateCheck.php
```diff
@@ -7,7 +7,10 @@ public function check() {
 			->selectRaw('orders.id')
 			->join('order_instalments', 'orders.id', '=', 'order_instalments.order_id')
 			->join('products', 'orders.product_id', '=', 'products.id')
+			->join('payment_method_product', 'orders.product_id', '=', 'payment_method_product.product_id')
+			->join('payment_methods', 'payment_method_product.payment_method_id', '=', 'payment_methods.id')
 			->whereRaw('order_instalments.due_date < products.signups_start')
+			->whereRaw('payment_methods.slug = \'instalments\'')
```
rogatty: I think a simple `where` would be enough.

PR #1423, opened 2019-04-12
PLAT-1057 | run browser tests before the relase

bognix:


docker-compose.yaml
```diff
@@ -96,15 +89,13 @@ services:
     ports:
       - "8755:8755"
     volumes:
-      - ${NFS_SOURCE_DIR}/wnl-platform/laravel-echo-server.json:/src/laravel-echo-server.json:ro
+      - ${LOCAL_CODE_PATH}/wnl-platform/laravel-echo-server.json:/src/laravel-echo-server.json:ro
```
rogatty: Was this intentional?

docker-compose.yaml
```diff
@@ -32,13 +32,6 @@ services:
     build:
       context: .
       dockerfile: Dockerfile.dev.php
-    environment:
-      # Required for browser tests
-      - APP_URL
-      - DEBUG_BAR
-      - SESSION_DOMAIN
-      # Required for ngrok
-      - P24_STATUS_URL
```
rogatty: We use ngrok for manual tests, too. I see no reason to remove this feature.

tests/Browser/Tests/Payment/Modules/MyOrdersModule.php
```diff
@@ -56,6 +56,7 @@ public function useCoupon(BethinkBrowser $browser, $value) {
 
 		$browser
 			->visit(new MyOrdersPage())
+			->waitFor('@discounts-tab', 30)
```
rogatty: I came up with another solution in https://github.com/bethinkpl/wnl-platform/pull/1424/files#diff-6942010429205f6d2bb25c26ef2ec9e4R60

tests/start-tests-env.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+DIR=$(dirname $0)
+cd ${DIR}/..
+
+docker-compose -f docker-compose.dusk.yml up -d
+
+printf "Wait 15 seconds for containers to boot...\n"
+sleep 15
```
rogatty: ☹️ can we write a script to check if the containers are running?

Dockerfile.dusk.php
```diff
@@ -0,0 +1,55 @@
...
+#
+
+FROM php:7.3.2-fpm-alpine3.9
+
+# Install PHP extensions
+RUN apk --no-cache add freetype-dev libjpeg-turbo-dev libpng-dev libzip-dev
+RUN docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ \
+&& docker-php-ext-install -j$(nproc) opcache bcmath gd zip mysqli pdo_mysql
+
+# Install New Relic Agent
```
rogatty: How about we skip it?

app/Providers/AppServiceProvider.php
```diff
@@ -146,7 +146,7 @@ protected function registerQueueLogger()
 
 	private function forceHttpsLinks() {
 		$url = app(UrlGenerator::class);
-		if (env('APP_ENV') !== 'dev') {
+		if (env('APP_ENV') !== 'dev' && env('APP_ENV') !== 'testing') {
```
rogatty: This should be reverted if we want to run tests using HTTPS.

PR #1427, opened 2019-04-15
[PLAT-1228] Create commands for slideshow export/import

gitkuba:


app/Console/Commands/SlideshowExport.php
```diff
@@ -0,0 +1,117 @@
...
+	 */
+	protected $signature = 'slideshow:export {screenId} {filename}';
+
+	/**
+	 * The console command description.
+	 *
+	 * @var string
+	 */
+	protected $description = '
+		Command that exports slideshow with all theirs connections including sides,
```
rogatty:
```suggestion
		Command that exports slideshow with all theirs connections including slides, 
```

app/Console/Commands/SlideshowImport.php
```diff
@@ -0,0 +1,191 @@
...
+
+	/**
+	 * @param array $subsections
+	 */
+	private function saveSubsections(array $subsections)
+	{
+		foreach ($subsections as $subsection) {
+			$originalSubsectionId = $subsection['id'];
+			unset($subsection['id']);
+			$subsection['section_id'] = $this->sections->get($subsection['section_id'])->id;
```
rogatty:
```
$this->sections->get($subsection['section_id'])->id
```
Does it get a section using its original id and returns the new id? If so, I'd add a comment or intermediate variable because it's not obvious.

app/Console/Commands/SlideshowImport.php
```diff
@@ -0,0 +1,191 @@
...
+		$this->subsections = collect();
+
+		$lesson = Lesson::find($this->argument('lessonId'));
+		if (is_null($lesson)) {
+			$this->error('Lesson not found');
+			return 1;
+		}
+
+		$filename = $this->argument('filename');
+		$fileContents = Storage::drive()->get($filename);
```
bognix: will it fail if file doesn't exist?

app/Console/Commands/SlideshowExport.php
```diff
@@ -0,0 +1,117 @@
...
+use App\Models\Section;
+use App\Models\Slideshow;
+use Illuminate\Console\Command;
+use App\Models\Screen;
+use App\Models\Subsection;
+use Illuminate\Support\Collection;
+use DB;
+use Storage;
+
+class SlideshowExport extends Command
```
bognix: did you consider adding unit tests to these two commands?

app/Console/Commands/SlideshowImport.php
```diff
@@ -0,0 +1,191 @@
...
+		$this->sections = collect();
+		$this->subsections = collect();
+
+		$lesson = Lesson::find($this->argument('lessonId'));
+		if (is_null($lesson)) {
+			$this->error('Lesson not found');
+			return 1;
+		}
+
+		$filename = $this->argument('filename');
```
bognix: Previous command you append the `.json` extension to the provided file name and in this command we expect the `.json` extension to be present in provided filename. Maybe it doesn't make sense but what do you think about unifying this?

bognix: Maybe I don't understand correctly how the Storage driver works but in previous command you save the file in exports directory. When passing path (filename) to this command to you need to specify the exports directory or it's magically resolved somewhere?

gitkuba: The thing is, I wanted to force the extension and export directory, so we don't create a mess in storage. On the other hand, I wanted to leave the free choice of import location and it seems to me more intuitive to pass full path with extension. I don't know. Maybe it lacks a better description. 🤷‍♂️

app/Console/Commands/SlideshowExport.php
gitkuba: Sure, how about we let this ticket to QA and I'll add some tests in the meantime?

.travis.yml
```diff
@@ -28,6 +28,7 @@ script:
   - php vendor/bin/phpunit --filter=Api
   - php vendor/bin/phpunit --filter=Http
   - php vendor/bin/phpunit --filter=Models
+  - php vendor/bin/phpunit --filter=Console
```
bognix: We already have the commands tests in `Unit/Commands` directory. Could you unify, either move your test there or move those test here.

tests/Console/SlideshowMigrationTest.php
```diff
@@ -0,0 +1,121 @@
...
+				'resources' => [
+					[
+						'name' => 'slideshows',
+						'id' => $originalSlideshow->id,
+					]
+				]
+			]
+		]);
+		$originalSlides = factory(Slide::class, self::SLIDES_COUNT)->create();
+		$originalSections = factory(Section::class, 10)->create([
```
rogatty:
```suggestion
		$originalSections = factory(Section::class, self::SECTIONS_COUNT)->create([
```

tests/Console/OrdersHandleUnpaidTest.php
```diff
@@ -21,6 +21,7 @@ class OrdersHandleUnpaidTestTest extends TestCase
 
 	public function testNoRemindersSent()
 	{
+		$this->markTestSkipped();
```
rogatty:
```suggestion
		$this->markTestSkipped('PLAT-1246');
```

lib/SlideParser/Parser.php
```diff
@@ -509,6 +511,9 @@ public function handleImages($html, bool $force = false, bool $resize = true)
 			$path = 'uploads/' . date('Y/m') . '/' . str_random(32) . '.' . $ext;
 			Storage::put('public/' . $path, $data, 'public');
 
+			if(strpos($imgTag, 'class="chart"') !== false) {
```
rogatty: Can we reuse https://github.com/bethinkpl/wnl-platform/pull/1427/commits/f05bfbeb59eaddaa968cc4a66c4e7cce0fd96c00#diff-48be8365dd82c0150ed9d5f27db7895aR40?

rogatty: Also, I'd move it next to `$template = self::IMAGE_TEMPLATE;`

gitkuba: ✅

PR #1429, opened 2019-04-15
PLAT-1167 | Stop making requests that we know will end up with 403

rogatty:
https://bethink.atlassian.net/browse/PLAT-1167

### DIFF
--- a/resources/assets/js/router.js
+++ b/resources/assets/js/router.js
@@ -39,16 +39,17 @@ import ModeratorsDashboard from 'js/components/moderators/ModeratorsDashboard';
 import MainUsers from 'js/components/users/MainUsers';
 import UserProfile from 'js/components/users/UserProfile';
 import Onboarding from 'js/components/onboarding/Onboarding';
+import SplashScreen from 'js/components/global/splashscreens/SplashScreen';
 
 Vue.use(Router);
 
 const routes = [
 	{
-		name: 'course',
 		path: '/app/courses/:courseId',
 		component: Course,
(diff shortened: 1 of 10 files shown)


resources/assets/js/router.js
```diff
@@ -39,16 +39,17 @@ import ModeratorsDashboard from 'js/components/moderators/ModeratorsDashboard';
 import MainUsers from 'js/components/users/MainUsers';
 import UserProfile from 'js/components/users/UserProfile';
 import Onboarding from 'js/components/onboarding/Onboarding';
+import SplashScreen from 'js/components/global/splashscreens/SplashScreen';
 
 Vue.use(Router);
 
 const routes = [
 	{
-		name: 'course',
```
kvas-damian: what was wrong with this `name`?

kvas-damian:
Navigation is missing on the splash screen page.

resources/assets/js/router.js
rogatty: It was causing a warning:
```
[vue-router] Named Route 'course' has a default child route. When navigating to this named route (:to="{name: 'course'"), the default child route will not be rendered. Remove the name from this route and use the name of the default child route for named links instead.
```
I checked that we don't navigate to `'course'` route anywhere.

rogatty:
>Navigation is missing on the splash screen page.
Fixed in https://github.com/bethinkpl/wnl-platform/pull/1429/commits/d974b8ab48f5203199ecac8ec4985625469ab72f

PR #1435, opened 2019-04-17
PLAT-1164 | Generate instalments on method or coupon change

kvas-damian:


app/Http/Controllers/Ajax/PaymentAjaxController.php
```diff
@@ -33,6 +33,8 @@ public function setPaymentMethod()
 
 		if ($method !== Order::PAYMENT_METHOD_INSTALMENTS) {
 			$order->orderInstalments()->delete();
+		} else {
+			$order->generateAndSavePaymentSchedule();
```
bognix: I would be more explicite here and compare `$method` with `Order::PAYMENT_METHOD_INSTALMENTS` to avoid surprises in the future if we add one more payment method.

rogatty: But we already do this 🤔 We could revert the condition but ultimately it works exactly the same, even if there will be an additional method.

PR #1436, opened 2019-04-17
 PLAT-1232 | Generate invoice only if order was not paid previously

rogatty:
https://bethink.atlassian.net/browse/PLAT-1232


lib/Invoice/Invoice.php
```diff
@@ -569,15 +574,6 @@ protected function getVatValue(Order $order)
 		return $order->product->vat_rate / 100;
 	}
 
-	private function advanceInvoiceSum()
```
rogatty: It's not used.

app/Jobs/OrderPaid.php
```diff
@@ -54,17 +62,23 @@ protected function handleCoupon()
...
 	protected function sendConfirmation()
 	{
-		\Log::notice("OrderPaid: sendConfirmation called for order #{$this->order->id}");
 		$order = $this->order;
 
-		\Log::debug('Issuing invoice and sending order confirmation.');
+		Log::notice("OrderPaid: sendConfirmation called for order #{$order->id}");
+
+		if (!$this->generateInvoice) {
+			Log::debug('Skipping invoice and order confirmation.');
```
bognix: this situations are quite rare and I think adding the order->id to the message and using the same log level as with remaining messages may be useful at some point.

rogatty: Well, it's the same log level and format as in `Log::debug('Issuing invoice and sending order confirmation.');`. Do you think we should update both? I don't know what was the reasoning behind choosing those levels.

PR #1438, opened 2019-04-17
PLAT-1239 | Use CORS middleware in products list API

rogatty:
Works fine in console on https://bethink.staging.wpengine.com/ and https://wiecejnizlek.pl/
```
let x = await fetch("https://lek.wiecejnizlek.pl/api/v1/products/current");

Access to fetch at 'https://lek.wiecejnizlek.pl/api/v1/products/current' from origin 'https://wiecejnizlek.pl' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

```
let x = await fetch("https://staging.wiecejnizlek.pl/api/v1/products/current");
let y = await x.json();
y.find(product => product.slug === 'wnl-online').quantity;

674
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,6 +5,7 @@
 	"type": "project",
 	"require": {
 		"php": ">=7.3.2",
+		"barryvdh/laravel-cors": "^0.11.3",
 		"barryvdh/laravel-dompdf": "^0.8.0",
 		"bethink/elastic-scout": "3.0.8",
 		"bschmitt/laravel-amqp": "^1.2",
@@ -20,8 +21,8 @@
 		"league/flysystem-aws-s3-v3": "^1.0",
 		"league/fractal": "^0.15.0",
 		"nao-pon/flysystem-google-drive": "~1.1",
-		"rap2hpoutre/fast-excel": "^1.1",
 		"predis/predis": "^1.1",
+		"rap2hpoutre/fast-excel": "^1.1",
 		"sentry/sentry": "^1.10",
 		"symfony/css-selector": "3.1.*",
 		"symfony/dom-crawler": "3.1.*"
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -33,7 +34,7 @@
 		"fzaninotto/faker": "~1.4",
 		"johnkary/phpunit-speedtrap": "^3.0",
 		"laravel/dusk": "^5.0",
-		"mockery/mockery": "^1.2",
+		"mockery/mockery": "0.9.11",
```
kvas-damian: Why did you downgrade this lib?

config/cors.php
```diff
@@ -0,0 +1,25 @@
...
+	'allowedOrigins' => [
+		'http://wiecejnizlek.pl',
+		'https://wiecejnizlek.pl',
+		'https://bethink.staging.wpengine.com'
+	],
+	'allowedOriginsPatterns' => [],
+	'allowedHeaders' => ['*'],
+	'allowedMethods' => ['*'],
+	'exposedHeaders' => [],
+	'maxAge' => 0,
```
kvas-damian: I believe we can cache preflight request response: https://github.com/asm89/stack-cors#options

composer.json
rogatty: We already use this version, see https://github.com/bethinkpl/wnl-platform/blob/fd4fbc3637a1f9993d0a8cecc50129ef27a30f40/composer.lock#L6395-L6396

I just locked it here so we can safely use composer CLI, e.g. I needed `composer require barryvdh/laravel-cors`. Without the lock, we're upgrading Mockery to new version and the tests throw:
```
testing.ERROR: No matching handler found for Mockery_1_Illuminate_Redis_RedisManager::set('UserState:Course:1:1:1', '"foo bar"'). Either the method was unexpected or its arguments matched no expected argument list for this method
```

PR #1439, opened 2019-04-17
Use queue:listen so we don't have to restart workers after code change

rogatty:
Quoting https://divinglaravel.com/queue-workers-how-they-work:

> The `queue:listen` command simply runs the `queue:work --once` command inside an infinite loop, this will cause the following:
> 
> - An instance of the app is booted up on every loop.
> - The assigned worker will pick a single job and execute it.
> - The worker process will be killed.
> 
> Using `queue:listen` ensures that a new instance of the app is created for every job, that means you don't have to manually restart the worker in case you made changes to your code, but also means more server resources will be consumed.


bognix:
merged to PLAT-1232

PR #1440, opened 2019-04-18
[PLAT-1245] Parser gif bug

preziu:


lib/SlideParser/Parser.php
```diff
@@ -520,7 +520,12 @@ public function handleImages($html)
 			}
 
 			$path = 'uploads/' . date('Y/m') . '/' . str_random(32) . '.' . $ext;
-			Storage::put('public/' . $path, $data->__toString(), 'public');
+			
+			if (is_string($data)) {
+				Storage::put('public/' . $path, $data, 'public');
+			} else {
+				Storage::put('public/' . $path, $data->__toString(), 'public');
```
rogatty: Let's not duplicate the code. How about this instead?
```
$data = is_string($data) ? $data : $data->__toString();
Storage::put('public/' . $path, $data, 'public');
```

preziu: sure!

rogatty:
Seems to be obsolete after https://github.com/bethinkpl/wnl-platform/pull/1427/commits/032d613362af44b327b8d663d50127780c099f43

gitkuba:
Fixed in #1427

PR #1441, opened 2019-04-19
PLAT-1229 | Fix saving lesson progress

kvas-damian:
We have to use right prop types to make that functions like `.includes()` or strict comparison operator works as expected.
Instead of applying `parseInt` here and there I decided to fix the root cause.


resources/assets/js/components/course/Lesson.vue
```diff
@@ -107,7 +107,28 @@ export default {
...
+			type: Number,
+			required: true,
+		},
+		lessonId: {
+			type: Number,
+			required: true,
+		},
+		screenId: {
+			type: Number,
+			required: true,
```
rogatty: `screenId` and `slide` are optional in the router.

resources/assets/js/router.js
```diff
@@ -63,13 +63,21 @@ const routes = [
...
+				}),
 				children: [
 					{
 						name: resource('screens'),
 						path: 'screens/:screenId/:slide?',
 						component: Screen,
-						props: true,
+						props: route => ({
+							screenId: Number(route.params.screenId),
+							slide: Number(route.params.slide),
```
rogatty: `slide` is optional based on the path.

resources/assets/js/router.js
```diff
@@ -63,13 +63,21 @@ const routes = [
 				name: resource('lessons'),
 				path: '/app/courses/:courseId/lessons/:lessonId',
 				component: Lesson,
-				props: true,
+				props: route => ({
+					courseId: Number(route.params.courseId),
+					lessonId: Number(route.params.lessonId),
+					...route.params.screenId && { screenId: Number(route.params.screenId) },
+					...route.params.slide && { slide: Number(route.params.slide) },
```
rogatty: This is a bit messy. The path doesn't include `screenId` or `slide`, so why are we passing them in?

kvas-damian: Should be better now.

resources/assets/js/router.js
```diff
@@ -61,25 +60,20 @@ const routes = [
...
-							screenId: Number(route.params.screenId),
-							slide: Number(route.params.slide),
-						})
-					}
-				],
+			},
+			{
+				// We need this route for backwards compatibility with progressStore in redis and notifications
+				name: resource('screens'),
+				path: '/screens',
```
rogatty: Don't we need the `screenId` and `slide` params too? Or are they passed to the redirect target under the hood?

kvas-damian: yes, they're

resources/assets/js/components/course/Lesson.vue
```diff
@@ -396,7 +425,7 @@ export default {
 				this.updateLessonNav({
 					activeSection: (this.currentSection && this.currentSection.id) || 0,
 					activeSubsection: parseInt(this.currentSubsection && this.currentSubsection.id,) || 0,
-					activeScreen: parseInt(this.screenId) || 0,
+					activeScreen: this.screenId || 0,
```
rogatty: Can we use `0` as a prop default instead?

resources/assets/js/services/progressStore.js
```diff
@@ -9,7 +11,11 @@ const setCourseProgress = ({ courseId, profileId }, value) => {
 	return axios.put(getApiUrl(`users/${profileId}/state/course/${courseId}`), value);
 };
 
-const setLessonProgress = ({ courseId, lessonId, profileId }, value) => {
+const setLessonProgress = ({ courseId, lessonId, profileId, route }, value) => {
+	if (route) {
+		// We need this to update last route not the previous one.
+		value.route = route;
```
rogatty: The comment isn't very helpful to me. Why does `value.route` store the previous route? Can we fix it higher up?

kvas-damian: I suspect someone swapped parameters order somewhere at the higher level that's why `route` attribute is wrong. 
I already made a few risky changes in this PR, that's why I didn't try to untangle that.

resources/assets/js/components/course/Lesson.vue
```diff
@@ -15,7 +15,11 @@
 							</div>
 						</div>
 					</div>
-					<router-view v-if="!isLessonLoading" @userEvent="onUserEvent" />
+					<wnl-screen
```
bognix: so now lesson can have only screens. It didn't have anything else in the past but it was possible :)

kvas-damian: For it isn't possible. For the future, we can change it. Although I feel it's more flexible to decide what do render at a Component level instead of a Route.

resources/assets/js/services/progressStore.js
rogatty: Let's put it in the comment.

resources/assets/js/components/course/Lesson.vue
```diff
@@ -322,13 +349,15 @@ export default {
...
+			} else {
+				this.updateLessonProgress();
 			}
 
 			this.updateLessonNav({
-				activeSection: (this.currentSection && this.currentSection.id) || 0,
-				activeSubsection: (this.currentSubsection && this.currentSection.id) || 0,
-				activeScreen: parseInt(this.screenId)
+				activeSection: (this.currentSection && this.currentSection.id) || null,
+				activeSubsection: (this.currentSubsection && this.currentSection.id) || null,
```
rogatty:
```suggestion
				activeSubsection: (this.currentSubsection && this.currentSubsection.id) || null,
```

resources/assets/js/components/course/Lesson.vue
```diff
@@ -394,9 +423,9 @@ export default {
 				}
 
 				this.updateLessonNav({
-					activeSection: (this.currentSection && this.currentSection.id) || 0,
-					activeSubsection: parseInt(this.currentSubsection && this.currentSubsection.id,) || 0,
-					activeScreen: parseInt(this.screenId) || 0,
+					activeSection: (this.currentSection && this.currentSection.id) || null,
+					activeSubsection: parseInt(this.currentSubsection && this.currentSubsection.id,) || null,
```
rogatty:
```suggestion
					activeSubsection: (this.currentSubsection && this.currentSubsection.id) || null,
```

app/Traits/EventContextTrait.php
```diff
@@ -51,7 +51,7 @@ protected function addEventContext(Model $model)
 					])->first()->order_number;
 
 					return [
-						'name' => 'screens',
+						'name' => 'lessons',
```
rogatty: This isn't currently used:
https://github.com/bethinkpl/wnl-platform/blob/7e34a4f6266c657617fbb3bd9b39725ede8affc5/resources/assets/js/components/course/screens/slideshow/Slideshow.vue#L296-L313
but I don't think it's a good time to do a clean up here.

PR #1442, opened 2019-04-19
PLAT-1238 | Lazy load slideshow media

rogatty:
https://bethink.atlassian.net/browse/PLAT-1238

When loading `/papi/v2/slideshow_builder/*` on MacBook Pro 2018 the average results are:

| Lesson | Load time on WiFi | Load time on Fast 3G | Data transferred |
| ------ | ---- | ------- | ---------------- |
| Kardiologia 1 | 2,76s -> 0,8s | 1.6min -> 3,9s | 23 MB -> 955 KB |
| Gastroenterologia 1 | 2s -> 0,81s | 25s -> 4,5s | 8,8 MB -> 1,1 MB |


app/Http/Controllers/Api/PrivateApi/SlideshowBuilderApiController.php
```diff
@@ -203,6 +203,10 @@ protected function getViewData($slides, $background)
 		$replace = ['<br>', ''];
 
 		$slidesContent = str_replace($search, $replace, $slides->implode('content', ' '));
+
+		// Force lazy loading
+		$slidesContent = preg_replace('/\ssrc="(.+)"/mU', ' data-src="$1"', $slidesContent);
```
bognix: I wonder if this shouldn't happen in `Parser::handleImages` and we should provide backward compatibility with already existing images.

kvas-damian: https://www.regular-expressions.info/modifiers.html 
> [...] It's use is strongly discouraged because it confuses the meaning of the standard quantifier syntax.

kvas-damian: Maybe we could use `(.+?)`?

resources/assets/js/slideshow.js
```diff
@@ -328,7 +316,3 @@ function toggleBookmark(parent) {
 function emitToggleFullscreen(state = true) {
 	parent.emit('toggle-fullscreen', state);
 }
-
-function refreshChart(index) {
```
bognix: this was done to make sure:
- charts are working in preview slide dialog (related slides in quiz questions and slides editor)
- charts are working after refreshing the slideshow using the "refresh slideshow" button that appears after admin edited the presentation user is currently on.

Could you make sure those two cases still work?

app/Http/Controllers/Api/PrivateApi/SlideshowBuilderApiController.php
kvas-damian: I think it's ok to store correct HTML in the DB. It's possible that we'll expose slides content in another context where lazy-loading should be disabled.

gitkuba: I was thinking the same, but I'd keep it that way for now. It's easy to rollback and I suppose it's a cheap operation and won't affect loading that much.

rogatty: I was thinking the same as @kvas-damian. The result of this method is cached, so it doesn't affect loading much.

resources/assets/js/slideshow.js
rogatty: It's broken at least in the admin panel.

rogatty: Fixed in https://github.com/bethinkpl/wnl-platform/pull/1442/commits/877a3cc388a72a8ad4cf622d04c5bde201aca521

PR #1443, opened 2019-04-19
Enable vue/require-prop-types and vue/require-default-prop

rogatty:
I started fixing stuff but then realized it's not doable in a reasonable time. Instead, I disabled the rules and added TODO comments for the old code. We'll have the new code linted while not being dragged by the old one.


resources/assets/js/components/messages/ConversationSnippet.vue
```diff
@@ -117,12 +117,13 @@ export default {
...
 	props: {
 		room: {
+			type: Object,
 			required: false,
 			default: () => ({})
 		},
 		profile: {
-			required: false,
 			type: Object,
+			required: false,
```
kvas-damian: not required

resources/assets/js/components/messages/ConversationsList.vue
```diff
@@ -131,6 +131,7 @@ export default {
 	},
 	props: {
 		withSearch: {
+			type: Boolean,
 			required: false,
```
kvas-damian: can be removed

resources/assets/js/components/quiz/SingleQuestion.vue
```diff
@@ -91,7 +91,7 @@ export default {
 	props: {
 		quizQuestionId: {
 			required: true,
-			type: [String, Number],
```
kvas-damian: `Invalid prop: type check failed for prop "quizQuestionId". Expected Number, got String.`

PR #1446, opened 2019-04-23
PLAT-1222 | Remove QnA from Overview.vue

kvas-damian:


resources/assets/js/store/modules/qna.js
```diff
@@ -15,14 +15,6 @@ function _resolveQuestion(questionId, status = true) {
 	});
 }
 
-function _getQuestionsLatest() {
-	return axios.get(getApiUrl('qna_questions/latest'), {
```
rogatty: I believe we can remove it from the API.

PR #1449, opened 2019-04-24
PLAT-1241 | Fix bodywork slides

rogatty:
https://bethink.atlassian.net/browse/PLAT-1241


webpack.mix.js
```diff
@@ -5,6 +5,8 @@ const sassOptions = {
 };
 
 mix.sass('resources/assets/sass/app.scss', 'public/css/app.css', sassOptions)
+	.sass('resources/vendor/reveal/reveal-theme-bodywork.scss', 'public/css/reveal-theme-bodywork.css', sassOptions)
```
rogatty: Unfortunately, it won't be rebuilt when running `yarn run hot`. I'd prefer to use a dynamic import in JS but it turned out to be a complicated thing to enable the last time I tried. These styles will most likely never change so I don't think it's worth the scope creep.

.travis.yml
```diff
@@ -25,10 +25,7 @@ before_script:
   - php artisan db:seed --class=TestSeeder
 
 script:
-  - php vendor/bin/phpunit --filter=Api
-  - php vendor/bin/phpunit --filter=Http
-  - php vendor/bin/phpunit --filter=Models
-  - php vendor/bin/phpunit --filter=Console
+  - php vendor/bin/phpunit --filter='Tests\\(Api|Console|Models|Unit)'
```
rogatty: We weren't running the tests in `Tests\Unit` directory, except the one in `Tests\Unit\Models`. I'd prefer to run **all** except the ones in `Tests\Browser` but I wasn't able to prepare a working pattern. The other solution is to add `@group` to all browser tests but I think this one is cleaner.

resources/vendor/reveal/reveal-theme-wnl.scss
```diff
@@ -0,0 +1,48 @@
+/* latin-ext */
+@font-face {
```
kvas-damian: Make sure you include light font-weight introduced in https://github.com/bethinkpl/wnl-platform/pull/1447

lib/SlideParser/Parser.php
```diff
@@ -37,18 +37,14 @@ class Parser
 
 	const PAGE_PATTERN = '/<p>((?!<p>)[\s\S])*(\d\/\d)[\s\S]*<\/p>/';
 
-	const IMAGE_PATTERN = '/<img.*src="(.*?)".*>/';
+	const IMAGE_PATTERN = '/<img.*?style="(.*?)".*?src="(.*?)".*?>/';
```
kvas-damian: what if there are no styles on the image tag?

lib/SlideParser/Parser.php
```diff
@@ -577,4 +572,30 @@ private function getJpg(\Intervention\Image\Image  $image): StreamInterface
 
 		return $canvas->insert($background)->stream('jpg', 80);
 	}
+
+	private function getImageHtml(array $match, string $mime, string $path): string
+	{
+		$class = null;
```
kvas-damian: It doesn't change anything, but I'd feel safer if we use an empty string as an initial value.

.travis.yml
bognix: To be honest I don't see much difference between your solution and previous one, as long as you have to whitelist all the directories the solution is not "perfect".

rogatty: The difference is that previously the pattern matched both directories and class names. Now it's more explicit by adding `Tests\\`. One line vs. multiple lines is just a minor optimization to run them all together.

lib/SlideParser/Parser.php
rogatty: From what I've seen there is at least `style=""` but we shouldn't depend on that, I'll fix it.

resources/views/course/slideshow.blade.php
```diff
@@ -22,6 +22,10 @@
 			border: 1px solid #7a7f91 !important;
 		}
 	</style>
+	@php
+		$themeStyleSheet = config('app.instance_name') === 'bodywork' ? 'reveal-theme-bodywork.css' : 'reveal-theme-wnl.css';
```
bognix: I would move this variable to controller.

resources/views/course/slideshow.blade.php
```diff
@@ -30,7 +34,7 @@
 			<img class="bookmark-image bookmark-image-remove" src="{{ asset('images/star.svg') }}" alt="Usuń z zakładek">
 			<span>zapisz</span>
 		</a>
-		<div class="theme-font-montserrat theme-color-white-blue" style="width: 100%; height: 100%;">
+		<div class="theme-font-{{ config('app.instance_name') === 'bodywork' ? 'overpass2' : 'montserrat' }} theme-color-white-blue" style="width: 100%; height: 100%;">
```
bognix: do we need this global class for fonts since we are loading different styles for each platform? We can use the same class and style it differently in each stylesheet.

resources/views/course/slideshow.blade.php
kvas-damian: Maybe we could store it in the `courses` tables?

rogatty: @kvas-damian how about we do that when there are more than 2 themes?

resources/views/course/slideshow.blade.php
rogatty: We can use a single class `theme-font` but we still need because we can't control the styles loading order when one is loaded through JS import and the other in the template. Dynamic imports would solve that.

See https://github.com/bethinkpl/wnl-platform/pull/1449/commits/a73f8652c327f0c047b3e9922b5607277d56fdb0

PR #1452, opened 2019-04-26
PLAT-1263 | Fix links in moderators feed

kvas-damian:


resources/assets/js/router.js
```diff
@@ -72,7 +72,7 @@ const routes = [
 			{
 				// We need this route for backwards compatibility with progressStore in redis and notifications
 				name: resource('screens'),
-				path: '/screens',
+				path: ':lessonId/screens/:screenId?/:slide?',
```
rogatty: Don't we need the `/app/courses/:courseId/lessons/` prefix?

kvas-damian: No, we don't:
* this is a child route so it inherits: `/app/courses/:courseId` from parent. 
* `lessons` was removed on purpose to not clash with existing `lessons` route

To be honest this route can be completely random - old urls are matched by `lessons` route. This one is only required to redirect `router-links` which generate url using path defined here. In most cases, the previous solution works as expected. The problem appears when you open a link in a new tab and there are no `params` in the `router-link` which are used for the redirect.

PR #1457, opened 2019-04-29
PLAT-1256 | use jest as unit tests framework

bognix:


resources/assets/js/components/global/Verify.spec.js
```diff
@@ -0,0 +1,30 @@
...
+				FileVerifiedSvg: '<div></div>',
+			},
+			sync: false
+		});
+	};
+
+	test('renders correctly when content not verified', () => {
+		const wrapper = createComponent();
+
+		expect(wrapper.vm.message).to.equal('Zweryfikuj');
```
kvas-damian: It should be visible for moderators. Do you define roles somewhere?

bognix: I didn't merge recent changes from PLAT-1231 here. After merging this branch this test should fail.

bognix: Wrong answer :D It's ok it's passing because the test is checking if the computed property message is correct when content is not verified. It's not checking if this message is actually visible.

Tests in this file are really naive and I'll update those later today. I just wanted to show you how it looks with `jest`.

jest.config.js
```diff
@@ -0,0 +1,13 @@
...
+		'^js/(.*)$': '<rootDir>/resources/assets/js/$1',
+		'^images/(.*)$': '<rootDir>/resources/assets/images/$1',
+	},
+	transform: {
+		'^.+\\.js$': 'babel-jest',
+		'.*\\.(vue)$': 'vue-jest',
+		'^.+\\.svg$': '<rootDir>/svgTransformer.js'
+	},
+	moduleFileExtensions: ['js', 'vue', 'json'],
+	setupFiles: ['./resources/assets/js/tests/setup.js']
```
bognix: check if I can use <rootDir> here.

jest.config.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+	moduleNameMapper: {
+		'^js/(.*)$': '<rootDir>/resources/assets/js/$1',
+		'^images/(.*)$': '<rootDir>/resources/assets/images/$1',
+	},
+	transform: {
+		'^.+\\.js$': 'babel-jest',
+		'.*\\.(vue)$': 'vue-jest',
+		'^.+\\.svg$': '<rootDir>/svgTransformer.js'
```
bognix: move the svgTransformer.js to assets/js/tests directory

rogatty:
Looks good. How about removing sinon and chai as they have built-in counterparts in jest?

jest.config.js
rogatty: Use what?

bognix: `<rootDir>`

bognix:
@rogatty makes sense. I'll merge this PR to PLAT-1231 and get rid of chai as part of working on PLAT-1231. In terms of sinon I don't have a lot of experience with `jest` equivalent but I think sinon is much more sophisticated.

PR #1458, opened 2019-04-29
PLAT-1261 | Handle no products

kvas-damian:


app/Http/Controllers/Auth/LoginController.php
```diff
@@ -46,14 +46,7 @@ public function __construct()
 	 */
 	public function showLoginForm()
 	{
-		$now = now();
-		$signupsOpen = App\Models\Product::select()
-			->where('slug', App\Models\Product::SLUG_WNL_ONLINE)
-			->whereDate('signups_start', '<=', $now)
-			->whereDate('signups_end', '>=', $now)
```
rogatty: Can't we add `->where('quantity', '>', 0)` instead of removing it?

kvas-damian: @adamkarminski wanted to limit the number of endpoints to the checkout. You can reach out to him if you have any concerns.

PR #1459, opened 2019-04-29
[PLAT-1247] Automatically mark content as verified

gitkuba:


kvas-damian:
In the ticket, I see the requirement "Exclude pages". I don't see it implemented.

rogatty:
Let's add some tests for those new events.

app/Observers/CommentObserver.php
```diff
@@ -5,16 +5,26 @@
...
 	{
-		//
+		$excludedDiscussions = Page::whereNotNull('discussion_id')->pluck('discussion_id')->toArray();
+		$excludedAnswers = QnaAnswer::select()
+			->whereHas('question', function ($query) use ($excludedDiscussions) {
+				$query->whereIn('discussion_id', $excludedDiscussions);
+			})->pluck('id')->toArray();
+		$excluded = in_array($comment->commentable_id, $excludedAnswers);
+
+		if($comment->user->isModerator() && !$excluded){
```
rogatty: I'd put `if (!$comment->user->isModerator()) { return; }` at the top to avoid all the unnecessary queries.

app/Observers/CommentObserver.php
```diff
@@ -5,16 +5,26 @@
...
-	public function created(Comment $comment)
+	public function creating(Comment $comment)
 	{
-		//
+		$excludedDiscussions = Page::whereNotNull('discussion_id')->pluck('discussion_id')->toArray();
+		$excludedAnswers = QnaAnswer::select()
+			->whereHas('question', function ($query) use ($excludedDiscussions) {
+				$query->whereIn('discussion_id', $excludedDiscussions);
+			})->pluck('id')->toArray();
+		$excluded = in_array($comment->commentable_id, $excludedAnswers);
```
rogatty:
```
select distinct commentable_type from comments;
+-------------------------+
| commentable_type        |
+-------------------------+
| App\Models\QnaAnswer    |
| App\Models\QuizQuestion |
| App\Models\Slide        |
+-------------------------+
```
Let's check if `$comment->commentable_type === QnaAnswer::class` before running any queries.

database/migrations/2019_04_29_105922_mark_content_verified.php
```diff
@@ -0,0 +1,52 @@
...
+	 *
+	 * @return void
+	 */
+	public function up()
+	{
+		$now = now();
+		$excludedDiscussions = Page::whereNotNull('discussion_id')->pluck('discussion_id')->toArray();
+		$excludedAnswers = QnaAnswer::whereIn('user_id', self::USERS_TO_MARK)
+			->whereDoesntHave('question', function ($query) use ($excludedDiscussions) {
+				$query->whereNotIn('discussion_id', $excludedDiscussions);
```
rogatty: Why not `whereIn` here and `whereHas` above? It's the same thing, but clearer, right?

app/Observers/CommentObserver.php
```diff
@@ -5,20 +5,42 @@
...
+use App\Models\QnaAnswer;
 use Illuminate\Foundation\Bus\DispatchesJobs;
 
 class CommentObserver
 {
 	use DispatchesJobs;
 
-
-	public function created(Comment $comment)
+	public function creating(Comment $comment)
```
bognix: why creating not created? I don't think we should block comments creation in case marking content fails.

app/Observers/CommentObserver.php
```diff
@@ -5,20 +5,42 @@
...
+		if ($comment->commentable_type !== QnaAnswer::class) {
+			$comment->verified_at = now();
+			return;
+		}
+
+		$excludedDiscussions = Page::whereNotNull('discussion_id')->pluck('discussion_id')->toArray();
+		$excludedAnswers = QnaAnswer::select()
+			->whereHas('question', function ($query) use ($excludedDiscussions) {
+				$query->whereIn('discussion_id', $excludedDiscussions);
+			})->pluck('id')->toArray();
```
kvas-damian: Result of this query can grow quickly. We'll be transferring a lot of data to php only to check if the id is included in the list. Can make that check in mysql instead?

PR #1463, opened 2019-05-02
PLAT-1274 | Handle products without delivery

kvas-damian:


app/Http/Forms/PersonalDataForm.php
```diff
@@ -69,13 +70,14 @@ public function buildForm()
 
 			->add('invoice', 'checkbox', [
 				'label' => trans('payment.invoice'),
+				'checked' => $forceInvoice
```
bognix: shouldn't it be read only if there is forceInvoice?

app/Http/Requests/Product/UpdateProduct.php
```diff
@@ -37,7 +37,8 @@ public function rules()
 		    'price' => 'numeric|required',
 		    'quantity' => 'integer|required',
 		    'initial' => 'integer|required',
-		    'delivery_date' => 'integer|required',
+		    'has_delivery' => 'boolean|required',
```
bognix: why it's required if it has a default value in DB?

resources/assets/js/admin/components/products/ProductDetailsEditor.vue
```diff
@@ -66,7 +67,12 @@
 						<small>Po tej dacie nie można już zapisać się na kurs.</small>
 				</wnl-datepicker>
 
-				<wnl-datepicker name="delivery_date" :config="datepickerConfig">
+				<wnl-check name="has_delivery" @input="onHasdeliveryChange">Wysyłka produktu</wnl-check>
```
bognix: `onHasDeliveryChange`

resources/assets/js/admin/components/products/ProductDetailsEditor.vue
```diff
@@ -161,6 +170,12 @@ export default {
 				this.$router.push({ name: 'product-edit', params: { id: data.id } });
 			}
 		},
+		onFormChange({ formData }) {
+			this.showDeliveryDate = formData.has_delivery;
+		},
+		onHasdeliveryChange( event) {
```
bognix:
```suggestion
		onHasdeliveryChange(event) {
```

app/Http/Requests/User/UpdateUserAddress.php
```diff
@@ -16,7 +16,7 @@ public function authorize()
 	{
 		$user = User::fetch($this->route('id'));
 
-		return $this->user()->can('update', $user->userAddress()->first());
+		return $this->user()->can('update', $user->userAddress()->firstOrNew([]));
```
bognix: why?

resources/assets/js/admin/components/products/ProductDetailsEditor.vue
```diff
@@ -66,7 +67,12 @@
 						<small>Po tej dacie nie można już zapisać się na kurs.</small>
 				</wnl-datepicker>
 
-				<wnl-datepicker name="delivery_date" :config="datepickerConfig">
+				<wnl-check name="has_delivery" @input="onHasdeliveryChange">Wysyłka produktu</wnl-check>
+				<wnl-datepicker
+					v-show="showDeliveryDate"
```
rogatty: Shouldn't this be `v-if`? We'll send the value to backend when it's hidden with CSS.

app/Http/Requests/User/UpdateUserAddress.php
kvas-damian: Now we allow users to not create the `userAddress`. Without this change, users can't fill in `userAddres` on the platform (Konto > Dane do wysyłki).

bognix: I don't think this is the best place to put this change. It's a side effect right now. If we want to ensure user always have userAddress (even empty) I would consider different place.

resources/assets/js/admin/components/products/ProductDetailsEditor.vue
kvas-damian: There is a problem with rendering datepicker behind `v-if`. It shows the `1970-01-01` instead of the provided date. Didn't want to fix that today - tight schedule.

app/Http/Requests/User/UpdateUserAddress.php
kvas-damian: What do you mean by a different place?
The clearest solution that comes to my mind is to create a separate endpoint for creation with separate validation etc. Is it something you had in mind?

app/Http/Requests/Product/UpdateProduct.php
kvas-damian: It's required because we use it in the code: https://github.com/bethinkpl/wnl-platform/blob/4523262a3cb4b8906ab9cf3d44d379e90483df2f/app/Http/Controllers/Api/PrivateApi/ProductsApiController.php#L116

app/Http/Requests/User/UpdateUserAddress.php
bognix: I was thinking about user observer, when you create a user we can create empty user address for this user.

PR #1475, opened 2019-05-14
PLAT-1287 | Update raven-js with sentry-browser

kvas-damian:
According to: https://ravenjs.com/ raven is deprecated.
Let's use `@sentry/browser` to see if errors are easier to track

Integration docs:
* https://docs.sentry.io/platforms/javascript/vue/
* https://forum.sentry.io/t/switching-to-sentry-javascript/4732


resources/assets/js/utils/logger.js
```diff
@@ -52,7 +58,10 @@ export default class Logger {
 	log(level, [message, extra = {}]) {
 		if (Logger.LEVELS[level] <= this.levelCode) {
 			if (this.useExternal(Logger.LEVELS[level])) {
-				Raven.captureMessage(message, { level, extra });
+				Sentry.withScope(scope => {
+					scope.setExtras(extra);
+					Sentry.captureMessage('test', 'info');
```
rogatty: Test?

kvas-damian: ups :)

PR #1477, opened 2019-05-15
PLAT-1252 | Parse annotations tag

bognix:


app/Models/ReferenceSlide.php
```diff
@@ -0,0 +1,24 @@
...
+	 * Get all of the owning referencable models.
+	 */
+	public function referencable()
+	{
+		return $this->morphTo();
+	}
+
+	public function slide()
+	{
+		return $this->belongsTo('App\\Models\\Slide');
```
rogatty: How about not using a string?
```suggestion
		return $this->belongsTo(App\Models\Slide::class);
```

app/Models/ReferenceSlide.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
+class ReferenceSlide extends Model
+{
+	protected $fillable = ['reference_color_id', 'reference_icon_id', 'slide_id', 'referencable_type', 'referencable_id'];
+	public $table = 'reference_slide';
```
kvas-damian: table name

app/Models/ReferenceSlide.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
+class ReferenceSlide extends Model
```
kvas-damian: maybe it could be a `Reference`?

lib/SlideParser/Parser.php
```diff
@@ -38,6 +40,8 @@ class Parser
 
 	const PAGE_PATTERN = '/<p>((?!<p>)[\s\S])*(\d\/\d)[\s\S]*<\/p>/';
 
+	const ANNOTATION_PATTERN = '/{a:(\d+):(\d+)(:(\d+))?}([\s\S]*){a}/';
```
kvas-damian: `([\s\S]*)` is too greedy:
`{a:1:1}asd{a}{a:2:2}zxc{a}` will be caught as one annotation

lib/SlideParser/Parser.php
```diff
@@ -607,4 +621,43 @@ private function getImageHtml(array $match, string $mime, string $path): string
...
+			\Log::warning("Slide with ID {$slide->id} has incorrect reference. Annotation with ID ${annotationId} does not exist");
+			$replacement = sprintf('<wnl-reference class="wnl-reference">%s</wnl-reference>', $referencedText);
+			$updatedContent = preg_replace(self::ANNOTATION_PATTERN, $replacement, $slide->content);
+			return $updatedContent;
+		}
+
+		$referenceData = [
+			'slide_id' => $slide->id,
+			'referencable_id' => $annotationId,
+			'referencable_type' => 'App\\Models\\Annotation',
```
kvas-damian: `Annotation::class`

lib/SlideParser/Parser.php
kvas-damian: To simplify handling matching results you could use non-capturing groups.

lib/SlideParser/Parser.php
```diff
@@ -607,4 +621,43 @@ private function getImageHtml(array $match, string $mime, string $path): string
...
 		return $html;
 	}
+
+	public function handleAnnotations($slide)
+	{
+		$match = $this->match(self::ANNOTATION_PATTERN, $slide->content);
+
+		$referenceColor = $match[0][1];
+		$annotationId = $match[0][2];
+		$referenceIcon = $match[0][4];
```
kvas-damian: `$referenceIconId`

lib/SlideParser/Parser.php
```diff
@@ -607,4 +621,43 @@ private function getImageHtml(array $match, string $mime, string $path): string
 
 		return $html;
 	}
+
+	public function handleAnnotations($slide)
+	{
+		$match = $this->match(self::ANNOTATION_PATTERN, $slide->content);
+
+		$referenceColor = $match[0][1];
```
kvas-damian: `$referenceColorId`

lib/SlideParser/Parser.php
```diff
@@ -607,4 +621,43 @@ private function getImageHtml(array $match, string $mime, string $path): string
...
+		}
+
+		if (!empty($referenceColor)) {
+			$referenceData['reference_color_id'] = $referenceColor;
+		}
+
+		$reference = ReferenceSlide::create($referenceData);
+
+		$replacement = sprintf('<wnl-reference class="wnl-reference" data-id="%s">%s</wnl-reference>', $reference->id, $referencedText);
+		$updatedContent = preg_replace(self::ANNOTATION_PATTERN, $replacement, $slide->content);
```
kvas-damian: What if there are 2 exactly the same annotation tags? Should we create 2 references or 1?

tests/Unit/lib/SlideParser/ParserTest.php
```diff
@@ -10,303 +15,401 @@
...
+//	<div class="sl-block" data-block-type="line" style="width: auto; height: auto; left: 82px; top: 663px;" data-block-id="c9e29b76eb8d310021c4511824ddcf12">
+//		<div class="sl-block-content" data-line-x1="-299" data-line-y1="-180" data-line-x2="498" data-line-y2="-180" data-line-color="#052c50" data-line-start-type="none" data-line-end-type="none" style="z-index: 14;" data-line-width="2px" data-line-style="dotted">
+//			<svg xmlns="http://www.w3.org/2000/svg" version="1.1" preserveaspectratio="xMidYMid" width="797" height="1" viewbox="-299 -180 797 1">
+//				<line stroke="rgba(0,0,0,0)" stroke-width="15" x1="-299" y1="-180" x2="498" y2="-180"></line>
+//				<line stroke="#052c50" stroke-width="2" stroke-dasharray="0 3.985" stroke-linecap="round" x1="-299" y1="-180" x2="498" y2="-180"></line>
+//			</svg>
+//		</div>
+//	</div>
+//</section>'
+//		];
```
kvas-damian: :see_no_evil:

app/Models/ReferenceSlide.php
rogatty: Now we have the `slide_id` column, so the name shouldn't be generic in case we want to use references outside the slides. However, can we use a two-way morph instead of `slide_id`? Then, this could be named `Reference`.

lib/SlideParser/Parser.php
rogatty: https://regex101.com/r/qOLcSB/1
```
{a:(\d+):(\d+)(?:(?::(\d+))|)}[^{]*{a}
```
Plus update `$referenceIcon = $match[0][4];` to use the 3rd group.

lib/SlideParser/Parser.php
```diff
@@ -346,6 +355,11 @@ public function downloadBackground($url)
 		return $fileName;
 	}
 
+	/**
+	 * @param $html
+	 * @return mixed|string|string[]|null
```
rogatty:
```suggestion
	 * @return string
```

lib/SlideParser/Parser.php
rogatty: I think we should iterate over `$match[i]` where each item represents one annotation.

lib/SlideParser/Parser.php
```diff
@@ -607,4 +621,43 @@ private function getImageHtml(array $match, string $mime, string $path): string
...
+		$referenceColor = $match[0][1];
+		$annotationId = $match[0][2];
+		$referenceIcon = $match[0][4];
+		$referencedText = $match[0][5];
+
+		if (empty($annotationId)) return $slide->content;
+
+		if (empty(Annotation::find($annotationId))) {
+			\Log::warning("Slide with ID {$slide->id} has incorrect reference. Annotation with ID ${annotationId} does not exist");
+			$replacement = sprintf('<wnl-reference class="wnl-reference">%s</wnl-reference>', $referencedText);
```
rogatty: Use `REFERENCE_PATTERN` const.

lib/SlideParser/Parser.php
```diff
@@ -607,4 +621,43 @@ private function getImageHtml(array $match, string $mime, string $path): string
...
+			$referenceData['reference_icon_id'] = $referenceIcon;
+		}
+
+		if (!empty($referenceColor)) {
+			$referenceData['reference_color_id'] = $referenceColor;
+		}
+
+		$reference = ReferenceSlide::create($referenceData);
+
+		$replacement = sprintf('<wnl-reference class="wnl-reference" data-id="%s">%s</wnl-reference>', $reference->id, $referencedText);
```
rogatty: Use `REFERENCE_TEMPLATE` const.

rogatty: Can we use `data-id` as a Vue component prop?

lib/SlideParser/Parser.php
rogatty: If the tags are the same, then I think there should be 1 reference used twice.

bognix: I think two because you may want to change one without changing the second.

lib/SlideParser/Parser.php
bognix: What exactly are you proposing? `reference-id=""`?

app/Models/ReferenceSlide.php
bognix: We'll have only the slide here for now, I don't really see a huge benefit of doing morph for "referencee" but I can add it. What about the name then? `referencee` sounds good?

rogatty: I don't think referencee is a word. How about `reference_source` and `reference_target`? Or do we need to have `...able` for Eloquent?

lib/SlideParser/Parser.php
rogatty: `reference-id` looks good but if `data-id` is a valid Vue prop then it's fine too, IMO.

app/Models/ReferenceSlide.php
bognix: By looking at the Laravel code I think we can change those names but it's something totally different than what we do now. It's not wrong, just different :)

bognix: By the way, referencee is a work

lib/SlideParser/Parser.php
bognix: @rogatty big thanks! I didn't know you speak regex ;)

tests/Unit/lib/SlideParser/ParserTest.php
bognix: I know I know, it's still WIP

app/Models/ReferenceSlide.php
bognix: I like the `reference_source` and `reference_target` idea, I'll update the code.

lib/SlideParser/Parser.php
bognix: `data-id` works well as the Vue prop.

lib/SlideParser/Parser.php
```diff
@@ -607,4 +627,55 @@ private function getImageHtml(array $match, string $mime, string $path): string
...
+
+				if (!empty($referenceColorId) && !empty(ReferenceColor::find($referenceColorId))) {
+					$referenceData['reference_color_id'] = $referenceColorId;
+				}
+
+				$reference = Reference::create($referenceData);
+				$replacement = sprintf(self::REFERENCE_FORMAT, $reference->id, $referencedText);
+			}
+
+			$updatedContent = str_replace($matchedTag, $replacement, $updatedContent);
```
kvas-damian: > str_replace — Replace all occurrences of the search string with the replacement string

To handle 2 unique references correctly you have to make sure this function replaces only one occurrence: https://stackoverflow.com/a/1252710

kvas-damian: Please add an unittest for this case

lib/SlideParser/Parser.php
```diff
@@ -607,4 +627,58 @@ private function getImageHtml(array $match, string $mime, string $path): string
 
 		return $html;
 	}
+
+	/**
+	 * @param Slide $slide
+	 * @return mixed
```
rogatty:
```suggestion
	 * @return string
```

tests/Unit/lib/SlideParser/ParserTest.php
```diff
@@ -307,6 +316,186 @@ public function cleanSlideProvider(): \Generator
...
 	}
+
+	public function testAnnotationsParserWithoutIcon() {
+		$color = ReferenceColor::create(['color' => '#000000', 'name' => 'foo']);
+		$annotation = Annotation::create(['title' => 'foo', 'description' => 'bar']);
+		$slideContent = sprintf('<section><span>{a:%s:%s}REDACTED{a}</span></section>', $color->id, $annotation->id);
+		$slide = Slide::create(['content' => $slideContent]);
+
+		$output = (new FakeParser)->handleAnnotations($slide);
+		$createdReference = Reference::latest()->first();
```
rogatty: This will get slower and slower if we don't use `->limit(1)`, right?

tests/Unit/lib/SlideParser/ParserTest.php
```diff
@@ -307,6 +316,186 @@ public function cleanSlideProvider(): \Generator
...
+
+		$this->assertEquals(sprintf('<section><span><wnl-reference class="wnl-reference" data-id="%s">REDACTED</wnl-reference></span></section>', $createdReference->id), $output);
+		$this->assertEquals( Annotation::class, $createdReference->reference_target_type);
+		$this->assertEquals( $annotation->id, $createdReference->reference_target_id);
+		$this->assertEquals( $color->id, $createdReference->reference_color_id);
+		$this->assertEquals( $icon->id, $createdReference->reference_icon_id);
+	}
+
+	public function testAnnotationsParserIfNoAnnotationFound() {
+		$createdReferences = Reference::count();
```
rogatty: This is a bit misleading, I'd rename it to `$existingReferences`.

tests/Unit/lib/SlideParser/ParserTest.php
```diff
@@ -307,6 +316,186 @@ public function cleanSlideProvider(): \Generator
...
+		$slide = Slide::create(['content' => $slideContent]);
+		$output = (new FakeParser)->handleAnnotations($slide);
+		$this->assertEquals($slideContent, $output);
+		$this->assertEquals($createdReferences, Reference::count());
+	}
+
+	public function testAnnotationsParserIfAnnotationDoesNotExist() {
+		$warningMock = Log::shouldReceive('warning')->once();
+		$createdReferences = Reference::count();
+		$existingAnnotation = Annotation::first();
```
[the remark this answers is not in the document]
rogatty: @kvas-damian proposed to use `$id = -1` but I'm not sure if that will work.

tests/Unit/lib/SlideParser/ParserTest.php
```diff
@@ -307,6 +316,186 @@ public function cleanSlideProvider(): \Generator
...
+			"<section><span><wnl-reference class=\"wnl-reference\">REDACTED</wnl-reference></span></section>",
+			$output
+		);
+		$this->assertEquals($createdReferences, Reference::count());
+		$warningMock->verify();
+	}
+
+	public function testAnnotationsParserIfReferenceColorDoesNotExist() {
+		$annotation = Annotation::create(['title' => 'foo', 'description' => 'bar']);
+		$existingReferenceColor = ReferenceColor::first();
```
rogatty: Same as https://github.com/bethinkpl/wnl-platform/pull/1477/files#r284722837

tests/Unit/lib/SlideParser/ParserTest.php
```diff
@@ -307,6 +316,186 @@ public function cleanSlideProvider(): \Generator
...
+		);
+		$this->assertNull($createdReference->reference_color_id);
+	}
+
+	public function testAnnotationsParserIfReferenceIconDoesNotExist() {
+		$color = ReferenceColor::create(['color' => '#000000', 'name' => 'foo']);
+		$annotation = Annotation::create(['title' => 'foo', 'description' => 'bar']);
+		$icon = ReferenceIcon::create(['fa_icon_name' => 'bar', 'name' => 'foo']);
+
+		if (empty($icon)) {
```
rogatty: It won't be empty, we just created it.

tests/Unit/lib/SlideParser/ParserTest.php
```diff
@@ -307,6 +316,186 @@ public function cleanSlideProvider(): \Generator
...
+		$this->assertNull($createdReference->reference_icon_id);
+	}
+
+	public function testAnnotationsParserIfMoreThanOneAnnotationOnSlide() {
+		$color = ReferenceColor::create(['color' => '#000000', 'name' => 'foo']);
+		$annotationOne = Annotation::create(['title' => 'foo', 'description' => 'bar']);
+		$annotationTwo = Annotation::create(['title' => 'fizz', 'description' => 'buzz']);
+
+		$slideContent = sprintf(
+			"<section><span>{a:%s:%s}FIRST{a}</span><span>{a:%s:%s}SECOND{a}</span></section>",
```
rogatty: Please add one more annotation after a newline.

tests/Unit/lib/SlideParser/ParserTest.php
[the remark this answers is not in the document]
bognix: no -1 is not cool, let's go with the solution you proposed.

lib/SlideParser/Parser.php
```diff
@@ -607,4 +630,61 @@ private function getImageHtml(array $match, string $mime, string $path): string
...
+				}
+
+				if (!empty($referenceColorId) && !empty(ReferenceColor::find($referenceColorId))) {
+					$referenceData['reference_color_id'] = $referenceColorId;
+				}
+
+				// This if is only to handle the legacy parser tags. In the past we didn't have two dimensions: colors and icons, only one.
+				// Legacy tags without declared icon and with color id equal to 4 should be handled as references without color but with image icon.
+				// This code can be removed once we know we won't import presentations from slides.com with legacy tags.
+				if ($referenceColorId == self::ANNOTATION_LEGACY_IMAGE_ID && empty($referenceIconId)) {
```
rogatty:
```suggestion
				if ($referenceColorId === self::ANNOTATION_LEGACY_IMAGE_ID && empty($referenceIconId)) {
```

rogatty: Also, I'd extract it to a separate function.

kvas-damian: It's a string - it won't work. You have to cast to int before.

rogatty: Right, let's make it explicit then:
```suggestion
				if ((int)$referenceColorId === self::ANNOTATION_LEGACY_IMAGE_ID && empty($referenceIconId)) {
```

PR #1487, opened 2019-05-20
Fix showing dismisibleAlert in non-vue files

kvas-damian:


resources/assets/js/utils/eventbus.js
```diff
@@ -0,0 +1,2 @@
+import Vue from 'vue';
+export const EventBus = new Vue();
```
rogatty: I would add a comment when to use it.

PR #1490, opened 2019-05-24
PLAT-1310 | Fix saving quiz questions in admin

kvas-damian:


resources/assets/js/components/global/form/Quill.vue
```diff
@@ -28,6 +28,7 @@
 
 	.quill-container
 		position: relative
+		margin-bottom: $margin-huge
```
rogatty: Is it used in QnA or something else user facing? If so, did you check how it will look?

PR #1493, opened 2019-05-30
[PLAT-1305] Open lessons at 2am

gitkuba:


resources/assets/js/components/user/plan/ManualPlanNodeItem.vue
```diff
@@ -34,7 +34,7 @@
 						:class="{'hasColorBackground': isEven}"
 						:value="startDate"
 						:config="datepickerConfig"
-						@onChange="(payload) => $emit('change', {newStartDate: payload, lesson: node.model})"
+						@onChange="(payload) => $emit('change', {payload, lesson: node.model})"
```
kvas-damian: `payload` means nothing for me. Can we use more explicit names when emitting events?

PR #1501, opened 2019-06-04
[PLAT-1335] Pass reset - don't show error for nonexistent email

gitkuba:


app/Http/Controllers/Auth/ForgotPasswordController.php
```diff
@@ -29,4 +31,35 @@ public function __construct()
...
+		// Fake a successful response to prevent user enumeration.
+		if($response == Password::INVALID_USER) {
+			usleep(mt_rand(5, 30)*100000);
+			return $this->sendResetLinkResponse($request, Password::RESET_LINK_SENT);
+		}
+		// Random sleep time is supposed to mess up response time in such a way, that
+		// success and failure responses are indistinguishable by their duration.
+		usleep(mt_rand(1, 10)*100000);
+
+		return $response == Password::RESET_LINK_SENT
```
rogatty: It can be either `Password::INVALID_USER` or `Password::RESET_LINK_SENT`, so I don't think we need this.

app/Http/Controllers/Auth/ForgotPasswordController.php
```diff
@@ -4,29 +4,58 @@
...
+	 */
+	public function sendResetLinkEmail(Request $request)
+	{
+		$start = microtime(true);
+		$this->validateEmail($request);
+
+		// We will send the password reset link to this user. Once we have attempted
+		// to send the link, we will examine the response then see the message we
+		// need to show to the user. Finally, we'll send out a proper response.
+		$response = $this->broker()->sendResetLink(
```
kvas-damian: `$response` is unused

app/Http/Controllers/Auth/ForgotPasswordController.php
```diff
@@ -4,29 +4,58 @@
...
+			$request->only('email')
+		);
```
kvas-damian: Maybe we could call the original method from the trait and ignore it's output? I'd avoid copying code when possible.

gitkuba: Calling original method has side effects on the `Response` object. Namely, it adds the message to it, we wanted to avoid showing 😬

PR #1502, opened 2019-06-04
PLAT-690 | Display References

bognix:
TODO:
- [x] Display references only for moderators
- [x] Add error handling
- [x] Add loading state when fetching the reference target content
- [ ] ~Vuex caching of reference targets~
- [x] Rethink the way we parse slides content
- [x] Fix unavailable getter isMobile in admin 

Some things were moved to https://bethink.atlassian.net/browse/PLAT-1255


resources/assets/js/components/course/screens/slideshow/SlideshowReveal.vue
```diff
@@ -212,15 +246,36 @@ export default {
 		];
 		this.keyboardCondition = (event) => keycodesWhitelist.indexOf(event.keyCode) > -1;
 	},
-	mounted() {
+	async mounted() {
 		this.updateBackgroundType();
-		this.setupReveal();
+		await this.setupReveal();
+		await Vue.nextTick();
```
rogatty: I don't think we need that. Reveal modifies classes but it shouldn't matter in this case.

resources/assets/js/components/course/screens/slideshow/SlideshowReveal.vue
```diff
@@ -212,15 +246,36 @@ export default {
...
+
+		if (this.$moderatorFeatures.isAllowed('access')) {
```
rogatty: Let's put this whole block in a method.

resources/assets/js/components/course/screens/slideshow/SlideshowReveal.vue
```diff
@@ -292,6 +328,28 @@ export default {
 				});
 			}
 		},
+		async initCurrentSlideAnnotations() {
+			if (this.$moderatorFeatures.isAllowed('access')) {
+				await Vue.nextTick();
+				const slide = this.slides[this.currentSlideIndex];
+
+				slide.references && slide.references.forEach(reference => {
+					const references = this.$refs.slides.querySelectorAll(`.stack.present wnl-reference[data-id="${reference.id}"]`);
```
rogatty: How about running `querySelectorAll` once and then filtering out incorrect ids? DOM access is slow.

resources/assets/js/admin/components/slides/references/SlideReferences.vue
```diff
@@ -36,10 +36,10 @@
 					<td>
 						<span
 							class="icon"
-							:style="{ color: reference.color ? reference.color.color : '#ccc' }"
-							:title="reference.color && reference.color.name"
+							:style="{ color: reference.color.color }"
```
bognix: color and icon names are always set?

resources/assets/js/components/course/screens/slideshow/SlideshowReveal.vue
```diff
@@ -1,33 +1,51 @@
...
+		<wnl-reference-modal
+			v-if="selectedReference"
+			:reference-id="selectedReference.id"
+			@closeModal="selectedReference = null"
+		>
+			<div slot="annotation-header" class="-stormGray reference-modal-annotation-header">
+				<span class="icon -cadetBlue is-small">
+					<i class="fa" :class="`fa-${selectedReference.icon.fa_icon_name}`" />
+				</span>
+				<strong>{{selectedReference.icon.name}}</strong>
```
bognix: why strong not CSS for styling?

resources/assets/js/components/course/screens/slideshow/SlideshowReveal.vue
```diff
@@ -64,6 +82,16 @@
 	.dark-custom-background
 		background: #0c1726 !important
 		color: white !important
+
+	.reference-color-icon
```
bognix: didn't we agree to use atomic by default for new styles? Or don't you want to mix atomic with "old style" component?

resources/assets/js/components/course/screens/slideshow/SlideshowReveal.vue
```diff
@@ -93,12 +121,18 @@
 
 <script>
 import $ from 'jquery';
-import { nextTick } from 'vue';
+import Vue from 'vue';
```
bognix: `import Vue, { nextTick } from vue`

resources/assets/js/admin/components/slides/references/SlideReferences.vue
kvas-damian: yes, we use one method for injecting color and icon into reference instance which includes defaults.

resources/assets/js/components/course/screens/slideshow/SlideshowReveal.vue
kvas-damian: why css not strong?

resources/assets/js/components/course/screens/slideshow/SlideshowReveal.vue
kvas-damian: I thought we agreed to not mix "old style" with atomic. I can change that if you think we should use atomic everywhere now.

resources/assets/js/components/course/screens/slideshow/SlideshowReveal.vue
bognix: because CSS is for styling and strong feels like styling leaking to markup.

resources/assets/js/components/course/screens/slideshow/SlideshowReveal.vue
bognix: Make sense, just wanted to double check.

bognix:
LGTM

resources/assets/js/components/global/modals/ReferenceModal.vue
```diff
@@ -0,0 +1,101 @@
...
+		};
+	},
+	computed: {
+		computedContent() {
+			if (this.content) return this.content;
+			return this.referenceContent;
+		},
+	},
+	async mounted() {
+		if (this.referenceId) {
```
gitkuba: Can be reversed as well.

resources/assets/js/components/global/annotations/Annotation.vue
```diff
@@ -0,0 +1,76 @@
...
+		};
+	},
+	async mounted() {
+		await nextTick();
+		if (!this.$refs.content) {
+			return;
+		}
+
+		this.images = [...this.$refs.content.querySelectorAll('img')];
+		this.images.forEach((image, index) => image.addEventListener('click', () => this.currentImageIndex = index));
```
rogatty: It would be safer to remove these event listeners. https://stackoverflow.com/a/12528067

resources/assets/js/components/global/modals/ReferenceModal.vue
```diff
@@ -0,0 +1,101 @@
...
+			this.isLoading = true;
+
+			try {
+				const { data } = await axios.get(getApiUrl(`references/${this.referenceId}`), {
+					params: {
```
gitkuba: Do we really need to pass params here since it's a "get"?

resources/assets/js/components/global/modals/ReferenceModal.vue
```diff
@@ -0,0 +1,101 @@
...
+						include: 'reference_target'
+					}
+				});
+				this.referenceContent = data.included.reference_targets[data.reference_target[0]].description;
```
gitkuba: Do we want to handle nonexistent property here?

rogatty: It's handled by try catch.

resources/assets/js/components/global/modals/ReferenceModal.vue
rogatty: Does it matter?

resources/assets/js/components/global/annotations/Annotation.vue
kvas-damian: I'm not sure. when annotation modal is closed, the component is removed so all references to these domNodes are removed. It suggests me that we don't have to manually remove event listeners. Those are just my feelings, we can test it.

rogatty: Let's test it in https://bethink.atlassian.net/browse/PLAT-1255

PR #1503, opened 2019-06-05
PLAT-1299 | Slides list referencing annotation

kvas-damian:


resources/assets/js/admin/components/slides/annotations/AnnotationsEditor.vue
```diff
@@ -230,6 +286,21 @@ export default {
...
+				this.referencingSlides = Object.values(included.references).map(reference => included.reference_sources[reference.reference_source[0]]);
+			} else {
+				this.referencingSlides = [];
+			}
+		},
+		async preview(slide) {
+			const { data } = await axios.get(getApiUrl(`slides/${slide.id}?include=slideshows`));
+			this.previewSlide = {
+				...data,
+				backgroundImage: Object.values(data.included.slideshows)[0].background_url
```
bognix: if slide is not in slideshow it will fail.

resources/assets/js/admin/components/slides/annotations/AnnotationsEditor.vue
```diff
@@ -85,37 +101,74 @@
 			:content="description"
 			@closeModal="isPreviewVisible = false"
 		/>
+
+		<wnl-slide-preview
+			v-if="previewSlide"
+			:slides="[previewSlide]"
```
rogatty: We could pass `referencingSlides` here and use `:current-slide-index` to control which one is visible.

PR #1507, opened 2019-06-05
PLAT-1343 | ability to set exam tag id through admin panel

bognix:


resources/assets/js/admin/components/lessons/edit/ScreensMetaEditorExams.vue
```diff
@@ -0,0 +1,34 @@
...
+			class="input"
+			placeholder="ID"
+			type="number"
+		>
+	</div>
+</template>
+
+<script>
+export default {
+	name: 'ScreensMetaEditorExams',
```
rogatty: No need for that.

PR #1516, opened 2019-06-07
PLAT-1367 Temporary hardcoding a message and creating a followup todo

adamkarminski:
* I wanted to remove misleading info that suggested, that `instalments` payment method is still available.
* Data about payment methods availability was not easily accessible in the Order.vue component.
* I decided to hardcode the message and created a follow-up ticket with a due date on 2019/08/31: https://bethink.atlassian.net/browse/PLAT-1369

Please review and let me know what you think.


bognix:
@adamkarminski please fix linter.

PR #1517, opened 2019-06-07
[PLAT-1327] Deal with timeouts

gitkuba:


gitkuba:
@bethinkpl/platform_devs So it seems reading the response from s3 will sporadically timeout when we ask for the public URL  of an asset. Building the URL ourselves seems to be some solution. Not a perfect one for sure, so I'm still looking for other options...

bognix:
@gitkuba do you know why it was happening only on 2 endpoints?

gitkuba:
Concerning the `/quiz_sets/{id}/stats` end-point, timeout seems to happen when we initialize an eloquent model. I can't really tell why. My best idea so far was to optimize that end-point, since it was initializing dozens or hundreds of thousands of models. Now we're making counting and grouping in mysql and it's also like 10x faster 😆

gitkuba:
> @gitkuba do you know why it was happening only on 2 endpoints?

@bognix It seems that those end-points are heavily using URL generator, so maybe we're hitting some rate limit on AWS. A good starting point for further debugging. Thanks for drawing my attention to it.

gitkuba:
Well, we call slideshow transformer for each slide. Even though we need to include one or two slideshows, we call each slideshow's transformer in order to find unique ones. And it kinda makes sense. Unfortunately, each time we call a slideshow transformer, we also request a public URL of its background from AWS.

app/Http/Controllers/Api/PrivateApi/QuizStatsApiController.php
```diff
@@ -14,18 +14,19 @@ public function __construct(Request $request)
 
 	public function get($id)
```
rogatty: Can you add a unit test and check that it passes before and after the change? A lot is going on here.

lib/Bethink/Bethink.php
```diff
@@ -11,7 +11,9 @@ public function getAssetPublicUrl($assetPath)
 		}
 
 		if(config('filesystems.default') === 's3') {
-			return \Storage::url('public/' . $assetPath);
+			$bucket = config('filesystems.disks.s3.bucket');
+			$region = config('filesystems.disks.s3.region');
+			return "https://{$bucket}.s3.{$region}.amazonaws.com/public/{$assetPath}";
```
rogatty: We could cache the results per `$assetPath` but let's first unblock our users and then improve it.

app/Http/Controllers/Api/PrivateApi/QuizStatsApiController.php
gitkuba: Seems like we already have one https://github.com/bethinkpl/wnl-platform/blob/master/tests/Api/Quiz/QuizStatsTest.php

lib/Bethink/Bethink.php
gitkuba: https://bethink.atlassian.net/browse/PLAT-1377

PR #1519, opened 2019-06-10
PLAT-1372 | Fix bookmarks on questions list

kvas-damian:
We store user's position on quiz questions and after page reload we show the same page. We were always loading reactions from page 1 and that caused errors. Instead, let's use the stored page number.


rogatty:
@kvas-damian how about a brief description why this change was needed?

kvas-damian:
In the ticket comments: 
```
Steps to reproduce:

Go to: http://platform.local/app/questions/list
Use questions pagination to switch questions page (make sure current question number is >30)
Refresh the page
```

PR #1520, opened 2019-06-10
PLAT-1373 | Stop failing when there is no reaction state

rogatty:
I consistently get this error but can reproduce it only locally and on my account:
```
[Vue warn]: Error in render: "TypeError: Cannot read property 'hasReacted' of undefined"

found in

---> <Watch> at resources/assets/js/components/global/reactions/Watch.vue
       <WnlCommentsList> at resources/assets/js/components/comments/CommentsList.vue
         <QuizQuestion> at resources/assets/js/components/quiz/QuizQuestion.vue
           <WnlActivateWithShortcutKey> at resources/assets/js/components/global/ActivateWithShortcutKey.vue
             <QuizList> at resources/assets/js/components/quiz/QuizList.vue
               <QuestionsTest> at resources/assets/js/components/questions/QuestionsTest.vue
                 <QuestionsTestBuilder> at resources/assets/js/components/questions/QuestionsTestBuilder.vue
                   <QuestionsSolving> at resources/assets/js/components/questions/QuestionsSolving.vue
                     <QuestionsList> at resources/assets/js/components/questions/QuestionsList.vue
                       <Questions> at resources/assets/js/components/questions/Questions.vue
                         <Root>
```

I don't see any user impact as the Like button works fine even if there is error, but I also don't see any downsides of this change. What do you think, should we merge it?


resources/assets/js/mixins/reaction.js
```diff
@@ -2,7 +2,29 @@ import { get } from 'lodash';
...
+			required: true,
+		},
+		reactableId: {
+			type: Number,
+			required: true,
+		},
+		state: {
+			type: Object,
+			// state can be undefined in hard to reproduce edge cases
+			required: false,
```
kvas-damian: in this case we should set a default state

resources/assets/js/mixins/reaction.js
```diff
@@ -12,10 +34,12 @@ export const reaction = {
 	},
 	computed: {
 		hasReacted() {
-			return this.state.hasReacted;
+			// state can be undefined in hard to reproduce edge cases
```
kvas-damian: I think we should at least track them to sentry

resources/assets/js/mixins/reaction.js
```diff
@@ -18,7 +18,7 @@ export const reaction = {
 		state: {
 			type: Object,
 			// state can be undefined in hard to reproduce edge cases
-			required: false,
+			default: () => {},
```
kvas-damian: isn't it returning `undefined` now?

rogatty: In those edge cases we get `undefined` passed from the parent but I don't think it's a good default.

kvas-damian:
```
const a = () => {}; console.log(a())
// undefined
const b = () => ({}); console.log(b())
// {}
```

rogatty: 🤦‍♂

PR #1526, opened 2019-06-12
PLAT-1385 | Return early if there is nothing to update

rogatty:
https://bethink.atlassian.net/browse/PLAT-1385


resources/assets/js/plugins/comments-subscriber.js
```diff
@@ -21,10 +21,13 @@ const CommentsSubscriber = {
 		};
 
 		const handleCommentPosted = ({ objects }) => {
+			if (!OBJECTS_TYPES_TO_VUEX_MAP[objects.type]) {
```
kvas-damian: Can we add a small comment why we're doing it for future readers?

PR #1533, opened 2019-06-14
PLAT-1373 | Unify spacings and some styles between Quill and displayed HTML

kvas-damian:


resources/assets/js/admin/components/quizes/edit/QuizQuestionsSetEditor.vue
```diff
@@ -37,6 +37,7 @@
 			</span>
 			<label class="label">Opis</label>
 			<wnl-quill
+				class="content"
 				ref="descriptionEditor"
```
rogatty:
```
/home/travis/build/bethinkpl/wnl-platform/resources/assets/js/admin/components/quizes/edit/QuizQuestionsSetEditor.vue
  41:5  error  Attribute "ref" should go before "class"  vue/attributes-order
```

database/migrations/2019_06_14_121958_cleanup_empty_lines_in_quiz_questions.php
```diff
@@ -0,0 +1,35 @@
...
+	 */
+	public function up()
+	{
+		\DB::statement("update quiz_questions set text = replace(text, '<p><br></p>', '');");
+		\DB::statement("update quiz_questions set text = replace(text, '<p class=\"ql-indent-1\"><br></p>', '');");
+		\DB::statement("update quiz_questions set text = replace(text, '<p class=\"ql-indent-2\"><br></p>', '');");
+		\DB::statement("update quiz_questions set explanation = replace(explanation, '<p><br></p>', '');");
+
+
+		\DB::statement("update comments set text = replace(text, '<p><br></p>', '');");
```
rogatty: @kvas-damian what about `qna_questions` and `qna_answers`?

database/migrations/2019_06_14_121958_cleanup_empty_lines_in_quiz_questions.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class CleanupEmptyLinesInQuizQuestions extends Migration
```
rogatty: The name is outdated.

database/migrations/2019_06_14_121958_cleanup_empty_lines_in_quiz_questions.php
rogatty: Ah, I see we override Bulma margins in: https://github.com/bethinkpl/wnl-platform/blob/4f7d602cf69c55aad16f53d71f87c7e6f259e7e3/resources/assets/styles/_styleguide.scss#L367-L378
Why don't we unify and clean up those too?

rogatty: Actually, `qna-question-content` class isn't used anymore so there are now large gaps.

rogatty: On `db-lek-1`:
```
mysql> select count(1) from qna_questions where text like '%<p><br></p>%';
+----------+
| count(1) |
+----------+
|       92 |
+----------+
1 row in set (0.00 sec)

mysql> select count(1) from qna_answers where text like '%<p><br></p>%';
+----------+
| count(1) |
+----------+
|      192 |
+----------+
1 row in set (0.00 sec)
```

lib/Bethink/Bethink.php
```diff
@@ -80,4 +80,18 @@ public function mbSubstrReplace($string, $replacement, $start, $length = null, $
 			return mb_substr_replace($string, $replacement, $start, $length, $encoding);
 		}
 	}
+
+	// https://www.php.net/manual/en/function.rtrim.php#95802
+	public function strrtrim(string $message, string $strip): string
```
kvas-damian: can we have some tests for it?

app/Models/QnaQuestion.php
```diff
@@ -57,6 +58,11 @@ public function reactions()
 		return $this->morphToMany('App\Models\Reaction', 'reactable');
 	}
 
+	public function getTextAttribute()
+	{
+		return Bethink::strrtrim($this->attributes['text'], '<p><br></p>');
```
kvas-damian: Instead of fixing it while we query data from the DB, maybe it's better to do that on save?

PR #1534, opened 2019-06-14
PLAT-930 | Use Quill handler to upload images

rogatty:
https://bethink.atlassian.net/browse/PLAT-930


resources/assets/js/quill/handlers.js
```diff
@@ -0,0 +1,77 @@
...
+	input.click();
+
+	input.onchange = () => {
+		const file = input.files[0];
+		upload(this.quill, file);
+	};
+}
+
+export async function lucidChartHandler() {
+	const lucidChartString = prompt('Podaj adres lub ID Lucid Charta');
```
kvas-damian: you don't handle cancel button correctly

resources/assets/js/quill/handlers.js
```diff
@@ -4,73 +4,65 @@ import Quill from 'quill';
...
 
-	quill.insertEmbed(index, 'loading-state', '', Quill.sources.API);
-
-	// No need to add an empty paragraph
-	if (!(quill.getLeaf(index + 1)[0] instanceof BreakBlot)) {
-		quill.insertEmbed(index + 1, 'break', '', Quill.sources.API);
+	if (
+		quill.getText(index)[0] !== '\n' ||
+		// Image is being inserted directly before another one
+		quill.getLeaf(index)[0] instanceof ImageBlot
```
kvas-damian: Isn't it the same condition as in line 23.? Here we comment it as an image before the cursor and there as after.

PR #1550, opened 2019-07-15
PLAT-1455 | Delete lesson statuses separately so that accountant can catch it

gitkuba:


rogatty:
Let's add a test that verifies accountant entry (maybe checking that an event was triggered is enough?).

